PR #51, opened 2026-02-22
Enable Ctrl+F find-in-page in markdown viewer

rutgersc:
> The prevent-default plugin was blocking all browser shortcuts including Ctrl+F. Configure it to allow the FIND shortcut so users can search within rendered markdown.

Not to spam you with too many PRs, but I thought this one could be useful for other users as well.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -340,7 +340,11 @@ pub fn run() {
             }
             let _ = app.get_webview_window("main").expect("no main window").set_focus();
         }))
-        .plugin(tauri_plugin_prevent_default::init())
+        .plugin(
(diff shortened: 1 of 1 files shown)


alecdotdev:
could you provide a screenshot (or gif) of the feature?

rutgersc:
image

It covers the top bar, but it functions as expected

rutgersc:
If only things were that easy. 

> The native browser find bar renders at the very top of the webview, overlapping your custom title bar. Since WebView2
  doesn't let you reposition it, I'll replace it with a custom find bar in the frontend.

The alternative that it generated was pretty sloppy, using `window.find`. Constantly losing focus and such.

PR #159, opened 2026-05-20
fix(viewer): make Full-Width toggle actually toggle width; center default at ~780px

andrey-golovko:
## Summary

The Full-Width toggle in `MarkdownViewer.svelte` is currently a visual no-op for the document body. The default and `.full-width` rules are byte-for-byte identical:

```css
/* before */
.markdown-body {
    padding: 50px clamp(24px, 5vw, 50px);
    max-width: 100%;
}
.markdown-body.full-width {
    padding: 50px clamp(24px, 5vw, 50px);   /* same */
    max-width: 100%;                         /* same */
}
```

The `isFullWidth` state only feeds the `isOverhanging` derivation that decides whether the TOC overlaps the content (`MarkdownViewer.svelte:183`) — it does not affect the body's width or padding.

A side effect is that on wide monitors the body fills the entire viewer pane, producing very long line lengths.

## Change

One-line CSS edit — the default `.markdown-body` now uses a centering padding:

```diff
- padding: 50px clamp(24px, 5vw, 50px);
+ padding: 50px max(24px, calc((100% - 780px) / 2));
```

- On wide screens, `(100% - 780px) / 2` produces symmetric gutters that visually center an ~780 px content column. The `780` matches the existing magic number already used by `isOverhanging` at `MarkdownViewer.svelte:183`, so TOC overhang behavior is consistent.
- On narrow screens, the `max(24px, …)` floor keeps a sensible minimum gutter (same 24 px as before).
- `.markdown-body.full-width` is unchanged — its existing `clamp(24px, 5vw, 50px)` padding is now a genuine wide alternative, so toggling Full-Width has a visible effect for the first time.

`.markdown-body` remains the scroll container, so the scrollbar continues to sit at the pane edge (centering via horizontal padding rather than `margin: auto`).

## Out of scope

- No JS/TS changes. `isFullWidth` state, localStorage persistence, and TOC overhang logic at `MarkdownViewer.svelte:180-187` are untouched.
- Content width is hard-coded to 780 px to match the existing constant. Happy to make it a setting in a follow-up if you'd prefer.

## Test plan

- [x] `npm run build` succeeds (Svelte/Vite, no new warnings on touched rules)
- [x] Local Tauri release build (`cargo tauri build`) succeeds; toggling Full-Width visibly switches between centered ~780 px column and full-width on a wide monitor
- [ ] Cross-check at narrow viewport widths (≤ ~830 px) — gutter falls back to 24 px floor as expected
- [ ] Cross-check with TOC visible (`settings.showToc`) — TOC overhang behavior unchanged

### DIFF
--- a/.planning/codebase/ARCHITECTURE.md
+++ b/.planning/codebase/ARCHITECTURE.md
@@ -1,0 +1,259 @@
+<!-- refreshed: 2026-05-20 -->
+# Architecture
+
+**Analysis Date:** 2026-05-20
+
+## System Overview
+
+```text
+┌─────────────────────────────────────────────────────────────────────────┐
+│                      SvelteKit SPA (WebView / Tauri)                     │
+├──────────────────────────┬──────────────────────────────────────────────┤
+│   src/routes/+page.svelte│  src/lib/MarkdownViewer.svelte               │
+│   (route entry, mounts   │  (main app shell — 3546 lines)               │
+│    MarkdownViewer)        │  - tab lifecycle, file I/O orchestration,    │
+│                           │    mode switching, event wiring              │
+├──────────────────────────┴──────────────────────────────────────────────┤
+│                         UI Components Layer                              │
+│  src/lib/components/                                                     │
+│  Editor · TitleBar · TabList · Tab · Settings · Modal · Toast           │
+│  FindBar · Toc · ContextMenu · UpdateDialog · ZoomOverlay · HomePage    │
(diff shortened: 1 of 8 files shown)


PathGao:
Thanks for the report and proposed fix. This behavior is already implemented in `master` by e956f29: the default Markdown body is constrained and centered at 880 px, while `.full-width` removes that cap and margin. Closing as superseded.

PR #168, opened 2026-06-02
Saving rendering

dayeggpi:
- Improvement: When in rendered markdown view with file that was never saved/created, control+s doesnt work (nothing happens). also added save/save as in menu.
- Improvement: allow user to switch to rendered view without saving/creating file : if file was created and user switches to rendered view, it would save file automatically or ask user to save file before (depending on "save automatically" option), no way to see rendered view until file is saved.
- Bug fix: "autosave ON + confirm before save ON" now triggers prompt to ask user if they want to save (every 1.5seconds)


PathGao:
@dayeggpi — all three of these have landed, though not from this branch. Thank you for the report; the third one in particular took some tracing to find the cause of.

**1. Ctrl+S does nothing in rendered view on a never-saved file, and there is no Save / Save As in the menu.**
Fixed in #365. Saving works from reading mode, the menu carries both entries, and the untitled case (`path === ''`) is included rather than falling through.

**2. No way to see the rendered view until the file is saved.**
Fixed in #421. The cause turned out not to be a save policy at all: leaving the editor called `loadMarkdown(tab.path)`, which **re-reads the file from disk**, so a dirty tab had to resolve a save decision first or the preview would show the wrong text. It now renders the buffer — which the untitled branch two lines below was already doing. The modal is gone entirely.

Three other things fell out of that read, which is worth recording since you were the one who noticed the coupling:
- leaving the editor on a large file took the 50 KB preview branch, **re-truncating a complete buffer** and refusing saves until a background read finished;
- `if (!success) return` kept the tab in edit mode when the write failed — and `saveContent` returns false *permanently* for a read-only file, so **reading mode was unreachable for that tab, forever**;
- `loadMarkdown` writes into the *active* tab, so toggling split view on a background tab would have pulled the active tab's content over.

Closing a tab and closing a window still ask. Those buffers are about to cease to exist; a view toggle's does not.

**3. autosave ON + confirm-before-save ON prompts every 1.5 seconds.**
Fixed by `49cb75b`, before this round. With `confirmBeforeSave` on, background auto-save is disabled outright rather than firing and then asking — the Settings label promises confirmation before each save, so a silent debounced write contradicts it either way.

Closing this as addressed. If any of the three still misbehaves on a build from `master`, please reopen — the second one in particular changed behaviour rather than just fixing a failure, and it is worth a second pair of eyes.

PR #210, opened 2026-07-12
fix: keep folded sections aligned after window resize

PathGao:
## Problem

After the window is resized (or moved between displays with different scale factors), expanded heading sections lose alignment with their content. Two visible forms:

- a section-sized **gap** between a section and the next heading, or
- the next heading **overlapping** the previous section's still-visible text.

Root cause: folding animates `grid-template-rows: 1fr ⇄ 0fr` on `.foldable-content-wrapper`. After content reflows (narrower window → paragraphs wrap taller), the WebView does not reliably recompute the `1fr` track, so the wrapper's layout height goes stale relative to its rendered content, while `overflow: visible` lets the content paint outside it.

## Fix

- Replace the fractional grid track with an explicit measured height: the wrapper animates `height: var(--fold-content-height)` (collapsed: `height: 0`).
- New `src/lib/utils/foldLayout.ts` — `observeFoldLayout(root)` attaches a single `ResizeObserver` to each expanded `.content-inner`, batches all writes into one animation frame, and measures innermost wrappers first so nested folds stay consistent.
- Reflow-driven writes are applied **without** animating (suppress the transition around the write, force one layout, restore it), so on resize the wrapper tracks its content instantly; the 0.25 s fold/unfold animation is fully preserved.
- `MarkdownViewer` starts observation after each preview render and disconnects before the preview HTML is replaced.

## Testing

- `scripts/foldLayout.test.ts` (new) + whole `scripts/*.test.ts` suite passes; `svelte-check` reports 0 errors/0 warnings.
- Verified in a bundled macOS build across two displays with different scaling: shrinking the window no longer overlaps the following section (previously the wrapper lagged ~340 ms behind the reflowed content on every resize), expanding/collapsing still animates.
- Repro fixture included: `samples/fold-resize-regression.md`.

### DIFF
--- a/samples/fold-resize-regression.md
+++ b/samples/fold-resize-regression.md
@@ -1,0 +1,32 @@
+# Fold resize regression
+
+## Overview
+
+This short section must remain directly above the next heading after the window moves between displays or changes width.
+
+> Wrapped blockquote text changes height when the preview width changes. It must remain in normal document flow without covering the next heading.
+
+```text
+Line one
+Line two
+Line three
+```
+
+## Workflow
+
+1. Resize the preview from wide to narrow.
+2. Move the window between displays with different scaling.
+3. Fold and unfold this section.
+
+### Nested section
+
+This nested content exercises the inner fold wrapper.
+
+| Column A | Column B |
+| --- | --- |
+| A long value that wraps at narrow widths | Another value |
+
+## Final section
+
+This heading must never overlap Workflow or be separated by a viewport-sized gap.
+
--- a/scripts/foldLayout.test.ts
+++ b/scripts/foldLayout.test.ts
@@ -22,6 +22,6 @@ test('fold wrapper animates an explicit measured height instead of a fractional 
 test('preview lifecycle starts and cleans up fold observation', () => {
 	const viewer = readFileSync('src/lib/MarkdownViewer.svelte', 'utf8');
 
-	assert.match(viewer, /observeFoldLayout\(markdownBody\)/);
+	assert.match(viewer, /observeFoldLayout\((?:markdownBody|body)\)/);
 	assert.match(viewer, /stopObservingFoldLayout\?\.\(\)/);
 });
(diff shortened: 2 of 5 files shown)


PathGao:
**Why this only manifests on large, heavily-sectioned documents** — supplementary analysis that may help review/testing.

The defect itself is size-independent: with an animatable `1fr` track on an auto-height container, the wrapper's track size must be re-resolved whenever descendant content reflows. That depends on the engine's incremental-layout invalidation chain (dirty-marking → upward propagation → track re-resolution), and under this combination (animatable `fr` × indefinite height × deep text reflow) a link in that chain can be silently dropped, leaving a stale track. (We did not pinpoint the exact dropped link in WebKit source; the symptom and failure class are confirmed.)

What scales with document size is the bug's **exposure**, via four conditions:

1. **Wrapper count** — every heading adds a grid container; one stale track shifts everything below it, and P(at least one goes stale) grows with n.
2. **Reflow magnitude** — long wrapping paragraphs change height dramatically on width change; small docs mis-position by invisible pixels, large docs by whole viewports.
3. **Reflow waves** — large files load in two phases (50 KB preview, then full content via `requestIdleCallback`) plus async rich-content rendering (mermaid/KaTeX/highlight); each wave re-tests the invalidation chain after tracks were resolved.
4. **Offscreen layout** — during resize most wrappers of a long document are outside the viewport, where layout shortcuts make a skipped re-resolution most likely (inferred, not instrumented).

This is also why the fix is exposure-independent: `ResizeObserver` is a spec-guaranteed post-layout observation of committed sizes — it fires regardless of viewport position or which wave caused the change, so the conditions above now only scale the number of correctly-handled notifications, not the failure probability.

Test suggestion: a large fixture (many `h2`/`h3` sections, long wrapping paragraphs, a table + code block) resized narrow↔wide reproduces the original behavior far more reliably than small samples.

alecdotdev (approved):

Issue #212, opened 2026-07-12
macOS: Finder file-open silently lost when a second app copy launches (single-instance forwards argv only)


PathGao:
**Describe the bug**

On macOS, double-clicking a markdown file can silently do nothing — no window, no error — when a **second copy** of Markpad ends up launching while another copy is already running. The file-open event is lost entirely.

This happens whenever two copies of the app exist (e.g. the user once ran Markpad straight from the DMG or `~/Downloads`, and also has a copy in `/Applications`) — or in any dev/test setup running a locally built bundle alongside the installed one.

**Mechanism**

macOS delivers Finder file-opens as an **Apple Event** (`odoc`) — the file path is never in `argv`. The sequence that eats it:

1. LaunchServices resolves the `.md` handler to a specific bundle path (say `/Applications/Markpad.app`). The *running* copy has a different bundle path, so LS considers the handler "not running" and launches a new process, addressing the Apple Event to it.
2. The new process starts and `tauri-plugin-single-instance` runs immediately: it detects the primary instance (same bundle identifier), forwards its **argv** — which contains no file path on macOS — and exits.
3. The Apple Event arrives for a process that is already dead. It is dropped; the primary instance only gets an argv-only ping (it focuses the window, opens nothing).

`Single Instance Args: ["/Applications/Markpad.app/Contents/MacOS/Markpad"]` in the logs confirms the forwarded argv carries no file.

**To Reproduce**
1. Have two copies of Markpad (e.g. one in `/Applications`, one anywhere else) sharing the bundle id.
2. Launch copy B and leave it running.
3. Double-click any `.md` file in Finder (LS targets copy A).
4. Copy A flashes into existence, forwards, exits; the file never opens anywhere.

**Expected behavior**

The file opens in the running instance (or the second instance keeps ownership long enough to receive and forward the actual file paths).

**Suggested fix**

In the single-instance handling on macOS, don't exit before the open-documents Apple Event has been received: capture the opened paths (Tauri's `RunEvent::Opened`) and forward *those* to the primary instance instead of raw argv — or delay the exit until the first event-loop turn so `Opened` can fire. (Upstream, this is arguably a `tauri-plugin-single-instance` limitation: argv-forwarding is complete on Windows/Linux but loses Finder opens on macOS.)

**Device**
 - OS: macOS (Darwin 25.x)
 - App Version: v2.6.11 (also reproduced against current master builds)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[...]

PR #214, opened 2026-07-13
feat: independent native windows with menu-based tab transfer

### LINKED ISSUE #74: Ability to open multiple Markpad windows
**Is your feature request related to a problem? Please describe.**
I frequently wish to open several Markdown documents side by side on different monitors.  However, when running Markpad on Windows 11 it isn't possible to have multiple Markpad windows open at the same time.

**Describe the solution you'd like**
I would like the ability to open multiple Markpad windows.

**Describe alternatives you've considered**
The alternative is using other markdown editors to have markdown documents open in other windows, but this is klunky and it would be much nicer to us Markpad (which is great!)

**Additional context**
Thanks for considering this feature request :-)

### LINKED ISSUE #125: Add ability to detach a tab into a new window
**Describe the solution you'd like**
I would like to be able to click and hold a tab and drag it into a new window

### LINKED ISSUE #213: Editor attributes the previous tab's content to a restored tab whose rawContent is missing (silent file corruption with auto-save)
**Describe the bug**

When a restored tab has no usable content (`rawContent` is `undefined`), switching to it does not swap the Monaco buffer — the editor keeps displaying the **previous** tab's content under the new tab's title. The next content-change event then writes that stale buffer into the wrong tab's store, and with auto-save enabled the wrong content is written to the wrong file on disk. This is silent data corruption.

**How we hit it**

We hit this in a version-mix scenario while working on #211: a v2.6.11 build restored a session snapshot written in the newer window-state-only format proposed there (which intentionally carries no `rawContent`). `restoreState` assigns the parsed tabs wholesale:

```js
restoreState(jsonBuffer) {
    const data = JSON.parse(jsonBuffer);
    if (data && Array.isArray(data.tabs)) {
        this.tabs = data.tabs;           // ← shape never validated
        ...
```

so every restored tab had `rawContent === undefined`. Observed live (screenshot-verified): a window with tabs `3.md` (disk content `3333`) and `4.md` (disk content `4444`) displayed **`4444` under the `3.md` tab** after a tab switch. One edit event + auto-save later, `3.md` on disk becomes `4444` — we saw exactly that in an earlier uninstrumented run ("both files ended up with the same content").

**Why it happens**

1. `restoreState` trusts the snapshot's shape blindly — no field validation, no version check (the editor never wrote a version field historically).
2. The editor's tab-switch sync has no guard for non-string content: with `content` undefined the `setValue` swap is skipped/throws, leaving the stale buffer visible under the new tab.
3. The `onDidChangeModelContent` handler attributes whatever the editor holds to `tabManager.activeTabId` *at event time*, so a stale buffer + any change event = misattributed content.

**Expected behavior**

A tab with missing/invalid restored content should display as empty (or reload from its `path`), never as another tab's content — and nothing should be auto-saved from a buffer that was never attributed to that tab by the user.

**Suggested hardening (all cheap)**

- Normalize in `restoreState`: default `rawContent`/`originalContent` to `''`, drop entries without the expected fields; ignore snapshots whose `version` is newer than the build understands.
- Guard the editor sync: treat non-string tab content as `''` before `setValue`.

Note: #211 side-steps the *cross-version* trigger by moving the new snapshot format to its own storage key (an older build simply finds nothing to restore after a downgrade), but the editor-side robustness gap is independent of that PR and worth fixing for any malformed snapshot.

**Device**
 - OS: macOS (Darwin 25.x)
 - App Version: v2.6.11

PathGao:
> **Stacked on #211 (and sharing #210's baseline) — please review only the last 4 commits.** This branch is based on the two in-flight PRs; the multiwindow work semantically depends on #211's window-state snapshot, per-tab close review, and untitled numbering. I will rebase onto master as soon as they land, which will reduce the diff to the 4 commits listed below.

Implements #74: multiple independent native Markpad windows. A tab moves to a new window via its context menu ("Move to New Window"), carrying its full state — unsaved content, edit/split mode, scroll, history — with no save prompt: moving is not closing.

## Why menu-move instead of Chrome-style tab dragging

An earlier drag-based design was abandoned: nearly all of its cost served the drag gesture itself (cross-display coordinate math, window-bounds hit testing, cancel-anytime transactionality). An explicit menu action delivers the same user value (two documents side by side) at a fraction of the complexity, and the transfer broker introduced here is deliberately shaped so drag can be added on top later.

## Architecture

Each Tauri window is an isolated WebView/JS context, so the existing singleton stores are naturally per-window — the frontend needed no store refactor at all. The real work was three shared resources that multiwindow exposes:

1. **Moving a tab between JS contexts** — an in-memory Rust broker. The source stages a serialized tab; the destination window is created with the transfer token embedded in its label (`window-<token>` — a URL query would 404 in the asset protocol); the destination claims the payload and the source deletes its tab **only after a targeted claim acknowledgement**. Any failure (timeout, window-creation error) cancels and leaves the source tab untouched. Payloads are strictly validated — a tab whose content fields are not strings is never constructed (see #213 for what stale/undefined buffers do).
2. **The shared window-state snapshot** — only the `main` window persists and restores it. Secondary labels are per-session, so their snapshots could never be restored anyway; letting N windows share one write slot meant the last window closed overwrote everyone else, and every new window restored the full previous tab set. Persistence now goes through Rust (`fs::write` behind an awaited invoke): `setItem` is an async message to the WebKit storage process and loses its flush race when the last window's close exits the process. The localStorage keys are read once for migration and then removed; a downgraded build starts a fresh session instead of misreading anything. The key/format is unchanged (`savedTabsDataV2`) — nothing about the stored bytes changed, so no new key.

   *Authority principle:* the disk file is the sole authority for document content. The persistence snapshot never carries content (it must survive a power-off gap during which the file may change elsewhere); the transfer payload may (it lives ~a second inside one process, with the source alive until acknowledged). Whether a content-carrying snapshot is safe depends on the time span it must survive, not where it is stored.
3. **Broadcast events and global state** — Tauri's `emit`/`window.emit` broadcast to every window, so e.g. "New File" from a tab context menu would create a tab in *all* windows. Per-window intents now use `emit_to` + window-bound listeners. The file watcher was a single global slot (any window toggling auto-reload killed every other window's watcher) and is now keyed per window label. OS file-opens (single-instance argv and macOS `RunEvent::Opened`) previously hardcoded the `main` window — and silently dropped the file once main was closed; they now deliver to the focused viewer window, falling back to any viewer (#193 fixed bringing the receiving window forward; this fixes choosing the receiver).

Smaller pieces: untitled tabs keep their title on move unless the destination already uses it (the title is the document's identity; per-window smallest-free numbering is unchanged); the reviewing window raises itself before its close-review dialogs (they are in-app modals and could sit hidden behind another window); the capability ACL now covers `window-*` labels; `tauri-plugin-window-state` maps all secondary labels onto one state entry.

## Out of scope

Cross-window tab dragging, "move to existing window", restoring the multi-window layout across restarts, and the macOS double-copy file-open swallow (#212). Two pre-existing issues noticed during QA are filed separately rather than fixed here: context menus are not mutually exclusive, and the document context menu opens over modal dialogs.

## Testing

- **Automated:** 16 new node tests for the transfer snapshot (round-trip fidelity incl. unsaved content, strict validation rejections, arrival titling), 4 Rust broker unit tests (stage/claim/cancel/eviction), all existing suites green, `svelte-check` clean. The `windowStateRestore` guard test was updated for the Rust write-through (same invariant — v2 snapshots invisible to legacy builds — new mechanism).
- **Manual (macOS):** dirty and untitled tab moves; both window-close orders restore main's tabs; file-open routing via `open -a`, second-instance argv, and with main closed; per-window auto-reload watchers (toggling one window's off leaves the other's alive); two windows running their close reviews concurrently; event isolation (context-menu New File / Cmd+W / menu-bar Open all act on their own window only).
- **Environment note:** utilities that force-quit "windowless" apps (e.g. Supercharge) terminate via a Cocoa-level Quit that bypasses every window-close path; persistence semantics are "state as of the last graceful close". Ad-hoc-signed dev bundles are invisible to such utilities' accessibility checks and get culled — excluded during QA.

Closes #125

Fixes #213.

Closes #74.


PathGao:
@alecdotdev Would you be open to supporting and merging this native multi-window feature?

I have been testing it in daily use for a while, including dirty/untitled tab transfer, independent close flows, file-open routing, and per-window reload behavior. I have also rebased it onto the current master and resolved the conflicts.

I have follow-up maintenance PRs ready for the reliability issues around closing, preview rendering, loading, and release gates. After the core multi-window behavior is stable, I plan to submit the componentization separately: extracting the document-session, window-session, and Rust window-runtime responsibilities without mixing that refactor into this feature PR.

If the direction fits the project, I am ready to maintain the follow-up work and address review feedback.

alecdotdev:
great stuff! looks good to me

PR #225, opened 2026-07-14
feat: move tabs between windows, merge all windows, and a carry shortcut

PathGao:
> **Stacked on #214 — please review only the last 2 commits.** Rebases onto master once #214 lands.

Completes #214's menu-based transfer model with organization between EXISTING windows:

- **"Move to <window>"** in the tab context menu lists every other window. Hovering an entry makes the target window identify itself — a brief accent ring and centered badge, no focus or z-order change — so entries never need guessing. Moving reuses the acknowledged broker path end to end (stage → offer → destination claims → source deletes on ack; a destination mid-close-review ignores the offer and the timeout rolls the source back).
- **"Merge All Windows Here"** (hamburger menu): every other window hands its tabs over one by one through the same acknowledged path, then closes itself; recreatable HOME tabs are dropped.
- **⌘⇧M** carries the active tab to the next window (creation order, cyclic) with focus following — pressing repeatedly walks the tab onward; with no other window it degrades to detach-to-new-window.
- Rust side: a window registry (each window's frontend pushes its display metadata — tab state lives per WebView) serving the menu and cycle order, plus `offer_tab_to_window` / `focus_window`.

New strings cover all 26 locales.


PathGao:
Superseded by clean stacked replacement #279, which rebuilds the functional scope without stale merge history.

PR #226, opened 2026-07-14
feat: window tags — named color chips, with pinned tags as reopenable sessions

PathGao:
> **Stacked on #214 and the move-between-windows PR — please review only the last 4 commits.**

Optional **window tags**: a Chrome-tab-group-style colored name chip next to the hamburger button (hidden until set via the hamburger's "Window Tag…"; click to edit, eight fixed colors). Once set, the tag is the window's identity everywhere windows need telling apart: "Move to <window>" labels, the identify flash (ring and badge take the tag color), and the native window title ("name — active tab") so Mission Control shows it too.

**Pinned tags are named sessions.** Right-clicking the chip offers Pin/Unpin (and rename/clear — previously the document context menu leaked through here). A pinned window saves its open file paths under the tag at every close (per-tab review, last-tab close, and app-exit paths all hook in, awaited so the write cannot race process exit). The HOME page lists pinned tags as colored cards that reopen the whole set with the tag applied.

**Pinning is the single gateway to persistence**: an unpinned tag is session decoration and dies with its window; a pinned tag alone survives restarts — one switch governs name, color, saved file set, HOME card, and restart survival. (Editing a pinned tag preserves the pinned state and renames its saved session.) The tag rides the v2 snapshot as an additive field older builds ignore.

New strings cover all 26 locales.


PathGao:
Superseded by clean stacked replacement #280, which rebuilds the functional scope without stale merge history.

PR #234, opened 2026-07-22
feat: redesign settings and add cross-platform transparency

cmdr-chara:
## Summary

- redesign the Settings window around one shared Markpad workbench visual system
- add searchable, draggable, and resizable settings with consistent native controls and live font previews
- add configurable application/editor toolbars with ordering, visibility, and placement controls
- add whole-window solid/translucent surfaces with persisted opacity, backdrop dimming, and blur preferences
- align the titlebar overflow and home menus, preserving translated labels and shortcut columns
- complete the new settings copy for every supported language

### Editor

Editor settings

### Appearance

Appearance settings

### Toolbars

Toolbar settings

## Platform behavior

- Windows uses the native blur effect when translucency and blur are enabled
- macOS uses the under-window background effect
- Linux keeps alpha transparency and backdrop dimming, while native blur stays disabled because support depends on the compositor
- solid mode remains fully painted by the web UI; the installer window remains opaque

## Validation

- `npm run test:workflows` — 33/33 passing
- `npm run build` — passing
- `cargo check --manifest-path src-tauri/Cargo.toml` — passing
- `git diff --check` — passing
- `npm run check` — no new diagnostics; still reports the existing unused `@ts-expect-error` in `vite.config.js` and the existing timer type mismatch in `MarkdownViewer.svelte`

## Notes

The appearance panel now uses the same group, row, control, focus, and spacing rules as every other settings category instead of maintaining a separate custom layout.


PathGao:
@cmdr-chara @alecdotdev The #271–#280 multi-window/session stack is now merged. Our next planned item is #263: configurable preview-content width, including width-dependent ToC layout and print/PDF coverage. Its Settings controls would touch the same surface as this PR.

Before starting that work, could you confirm whether #234 is still intended to merge in its current form, or whether it needs revisions? We will keep #263 out of `Settings.svelte` until that direction is clear, then build it as a focused follow-up rather than create a large conflicting rebase.

PathGao:
Thanks for the detailed redesign proposal. The maintainer has decided not to adopt this redesign in its current form. The affected Settings surface will continue through smaller focused changes, so this broad and now-conflicting PR is being closed.

PR #237, opened 2026-07-24
fix: acknowledge tab transfer after destination render

PathGao:
Fixes the cross-window transfer loss window identified in #214. The broker retains the staged payload until the destination validates, inserts, and renders the tab, then acknowledges the source.\n\nValidation:\n- node --test --import tsx scripts/tabTransfer.test.ts\n- npm run check


PathGao:
Closing as duplicate: this rebased fix branch now updates PR #214 directly, so #214 contains the same head commit.

PR #243, opened 2026-07-24
fix: confirm dirty tabs before batch close

### LINKED ISSUE #242: fix: preserve unsaved tabs in close-other and close-right commands
The tab context actions for Close Others and Close to the Right call tabManager.closeTab directly. Dirty tabs are discarded without the existing save/discard/cancel flow.\n\nScope: close targets in tab-strip order through canCloseTab; stop immediately on cancellation or save failure; preserve the existing last-tab window-close behavior.

PathGao:
Fixes #242.

## Summary

- Route Close Others and Close to the Right through the existing save/discard/cancel flow.
- Stop the batch immediately when a user cancels or a save fails, preserving all remaining tabs.

## Validation

- `node --test --import tsx scripts/batchCloseConfirmation.test.ts`
- `npm run check`

### DIFF
--- a/scripts/batchCloseConfirmation.test.ts
+++ b/scripts/batchCloseConfirmation.test.ts
@@ -12,4 +12,6 @@ test('batch tab-close commands use the existing dirty-tab confirmation flow', ()
 	);
 	assert.match(viewer, /menu-tab-close-others[\s\S]*await closeTabsWithConfirmation\(tabsToClose\)/);
 	assert.match(viewer, /menu-tab-close-right[\s\S]*await closeTabsWithConfirmation\(tabsToClose\)/);
+	assert.match(viewer, /appWindow\.listen\('menu-tab-close-others', async \(event\)/);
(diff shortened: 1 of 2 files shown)


alecdotdev:
i think a prior PR may have created a conflict. just at work but ready to merge when resolved

PathGao:
@alecdotdev Thanks — I rebased #243 onto the current master and resolved the conflict while preserving the window-targeted listeners from #214. Its CI is running now.

Suggested merge order:
1. #241 — already clean, with the full platform build matrix passing.
2. #243 — once this current CI run passes.
3. I will then rebase and submit the remaining independent reliability fixes one at a time: #249 (path/input hardening), #254 (stale preview renders), #258 (viewer disposal), and #260 (interrupted restore recovery).

I would hold #225 and #226 for now. They are larger feature expansions still conflicted on the old #214 stack; after the independent fixes, I plan to do the no-behavior-change session/runtime componentization separately, then rebuild or rebase those features on the stable boundary.

PR #254, opened 2026-07-24
fix: ignore stale preview renders

### LINKED ISSUE #253: fix: prevent stale preview renders from replacing newer tab content
The split/edit preview debounce invokes render_markdown asynchronously, then writes its result without confirming that the active tab and raw content still match the request. A delayed render can replace a newer edit or a different active tab.\n\nGive each preview render request a revision, cancel pending timers in the effect cleanup, and discard stale completions.

PathGao:
Fixes #253.

## Summary

- Give each split/edit preview request a revision.
- Cancel its debounce timer when the reactive effect is invalidated.
- Apply a completed render only when the same active tab and raw content still match the request.

## Validation

- node --test --import tsx scripts/previewRenderRevision.test.ts
- npm run check

### DIFF
--- a/scripts/previewRenderRevision.test.ts
+++ b/scripts/previewRenderRevision.test.ts
@@ -1,0 +1,17 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const viewer = readFileSync('src/lib/MarkdownViewer.svelte', 'utf8');
+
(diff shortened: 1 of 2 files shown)


PathGao:
@alecdotdev This branch has been rebased onto the current `master` and force-pushed; the merge conflict is resolved. Validation passed: `node --test --import tsx scripts/previewRenderRevision.test.ts` and `npm run check`. This is the first PR in the dependency chain; please merge it before #258, #260, and #266.

PR #258, opened 2026-07-24
fix: guard viewer initialization after disposal

PathGao:
Fixes #257.

## Summary

- Mark `MarkdownViewer` initialization as disposed during teardown.
- Stop deferred syntax/math imports, session restore, initial file loading, and startup-path handling once disposed.
- Release listeners whose registration completed after the component was destroyed.

## Validation

- node --test --import tsx scripts/viewerDisposal.test.ts
- npm run check

PR #260, opened 2026-07-24
fix: recover from interrupted session restores

### LINKED ISSUE #259: fix: recover from an interrupted restored session
A Markdown file that hangs rendering during startup can leave its tab snapshot restored on every later launch, making the app unusable.\n\nRecord when automatic session restoration begins. If the previous restore was interrupted, discard only the saved tab snapshot on the next launch and open normally. Clear the marker after a successful restore; do not delete user files.

PathGao:
Fixes #259.

## Summary

- Mark automatic session restoration as in progress before restored tabs are read and rendered.
- Clear the marker after normal completion or a handled restore failure.
- On the next startup after an interruption, discard only the saved-tab snapshot and start normally; user documents are untouched.

## Validation

- node --test --import tsx scripts/interruptedSessionRestore.test.ts
- npm run check

### DIFF
--- a/scripts/interruptedSessionRestore.test.ts
+++ b/scripts/interruptedSessionRestore.test.ts
@@ -1,0 +1,18 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const viewer = readFileSync('src/lib/MarkdownViewer.svelte', 'utf8');
+
+test('session restore records work in progress before restoring tabs', () => {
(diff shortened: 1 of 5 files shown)


PathGao:
@alecdotdev This branch has been rebased onto the latest #258 branch and force-pushed; the merge conflict is resolved. Validation passed: `node --test --import tsx scripts/interruptedSessionRestore.test.ts scripts/windowStateRestore.test.ts` and `npm run check`. Please merge in this order: #254 → #258 → #260 → #266.

PR #249, opened 2026-07-24
fix: constrain theme and image file paths

PathGao:
Fixes #248.
Fixes #250.

## Summary

- Reject traversal, separators, and absolute paths in theme names and image path components.
- Canonicalize image directories after creation and reject symlinked directories outside the document parent.
- Prevent existing image symlinks from escaping the document parent.
- Use atomic writes for application theme state, downloaded themes, and decoded image data.
- Reject failed VSIX downloads and bound download size, archive entry count, expanded size, and manifest/theme JSON sizes.

## Validation

- project-local Cargo test: 4 passing Rust tests

PR #269, opened 2026-07-29
Revert "fix: constrain theme and image file paths"

alecdotdev:
Reverts alecdotdev/Markpad#249

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -113,73 +113,6 @@ fn atomic_write(target: &Path, bytes: &[u8]) -> std::io::Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


PathGao:
@alecdotdev Could you share the reason for reverting #249? I am preparing a smaller follow-up that preserves nested relative image directories while restoring the path and VSIX input boundaries. Knowing the concrete regression or compatibility concern will let me avoid repeating it.

alecdotdev:
@PathGao my apologies, I was merging the PRs in the sequence described here https://github.com/alecdotdev/Markpad/pull/249#issuecomment-5100473996 and only then noticed that #265 should have been merged first, so I reverted. No issues with the PR itself, if you make another PR with the changes let me know and I'll review and merge. 

Thanks again for all your work

Issue #250, opened 2026-07-24
security: bound VSIX theme downloads and archive expansion


PathGao:
The VSIX theme importer downloads and expands an archive from an external source without size or entry-count limits.\n\nAdd an HTTP status check, a streaming download cap, archive entry and uncompressed-size caps, and bounded manifest/theme JSON reads.

Issue #248, opened 2026-07-24
security: reject path traversal in theme and image file commands


PathGao:
Theme read/delete names and image save directories/filenames accept caller-controlled path components. A crafted Tauri command can escape the themes or document directory, including through a pre-existing symlinked image directory.\n\nScope: accept only a single safe path component for names/directories, canonicalize image destinations after creation, require they remain under the document parent, and add Rust tests for traversal, absolute paths, separators, and symlink escape.

PR #270, opened 2026-07-29
fix: restore file-command and VSIX input boundaries

PathGao:
## Summary

- Restore the theme and image command path boundaries from #249 after #269 reverted it solely to correct the merge order with #265.
- Bound VSIX HTTP downloads, archive expansion, and JSON reads.
- Keep the #265 UTF-8 fenced-code regression test alongside the restored path tests.

## Context

#249 was merged before #265 and then reverted by #269. @alecdotdev confirmed that the implementation itself had no issue and requested a new PR after #265.

Related: #248, #250, #249, #269.

## Validation

- `npm run check`
- `npm test` (117 passing)
- `cargo test` (20 passing)

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -113,6 +113,73 @@ fn atomic_write(target: &Path, bytes: &[u8]) -> std::io::Result<()> {
     Ok(())
 }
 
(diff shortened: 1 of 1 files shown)


PathGao:
@alecdotdev Reopened as requested on top of #265. This restores #249 unchanged except for retaining #265's UTF-8 fence regression test; all local checks are passing.

PR #274, opened 2026-07-29
refactor: extract document loading session

PathGao:
## Summary

- move document loading, tab selection, preview loading, and large-file completion guards into documentSession
- keep rendering, viewport measurement, recent-file UI, and error presentation as MarkdownViewer callbacks
- preserve the clean revision, dirty-state, and view-mode guard before a deferred full-file result can apply

Depends on #273 (and #272/#271). Merge after #273. This starts the document-session extraction; save, task mutation, watcher, and close flows remain separate follow-up steps.

## Validation

- npm run check
- npm test (117 passing)
- cargo test (18 passing)

PR #271, opened 2026-07-29
refactor: extract window runtime

PathGao:
Fixes #236.

## Summary

- Move process-wide window state, file watchers, window-state persistence, secondary-window creation, single-instance delivery, and window event handling into `window_runtime.rs`.
- Keep Tauri command names, payloads, builder wiring, and frontend behavior unchanged.
- Leave `lib.rs` responsible for top-level builder and command registration.

## Merge order

This is the first PR in the replacement chain:
1. window runtime extraction (this PR)
2. window session extraction
3. document session extraction
4. replacement for #225
5. replacement for #226

## Validation

- `npm run check`
- `npm test` (117 passing)
- `cargo test` (18 passing)

PR #275, opened 2026-07-29
refactor: move document saving into session

PathGao:
## Summary

- move normal and first-save document persistence into documentSession
- preserve snapshot-based dirty-state handling during asynchronous writes
- retain watcher suppression and recent-file presentation through explicit viewer callbacks

Depends on #274. Merge after #274. Save-as, task mutation, watcher, and close flows remain the next document-session steps.

## Validation

- npm run check
- npm test (117 passing)
- cargo test (18 passing)

### DIFF
--- a/scripts/windowClosePerTab.test.ts
+++ b/scripts/windowClosePerTab.test.ts
@@ -1,8 +1,10 @@
 import assert from 'node:assert/strict';
-import { readFileSync } from 'node:fs';
+import { existsSync, readFileSync } from 'node:fs';
 import test from 'node:test';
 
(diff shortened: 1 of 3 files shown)


PathGao:
@alecdotdev This follows #274. It moves normal and first-save persistence into documentSession while retaining the existing snapshot/TOCTOU and watcher-suppression invariants. The next dependent step covers save-as, task mutation, watcher, and close flows.

PathGao:
@alecdotdev Rebased after #271 merged; its direct parent is the updated #274 branch.

PR #276, opened 2026-07-30
refactor: move task updates into document session

PathGao:
## Summary

- move task-checkbox Markdown mutation and persistence into documentSession
- keep DOM checkbox discovery and visual state in MarkdownViewer
- preserve the in-memory update before save invariant

Depends on #275. Merge after #275. Watcher and close-confirmation ownership remain subsequent focused steps.

## Validation

- npm run check
- npm test (118 passing)
- cargo test (20 passing on the rebased parent)

PR #277, opened 2026-07-30
refactor: move watcher suppression into document session

PathGao:
## Summary

- move self-write grace tracking and watcher reload decisions into documentSession
- keep the window event subscription and actual document reload in MarkdownViewer
- preserve expired-marker cleanup before a real external reload

Depends on #276. Merge after #276. Close-confirmation ownership remains the final document-session follow-up before rebuilding #225.

## Validation

- npm run check
- npm test (119 passing)

### DIFF
--- a/scripts/documentWatcherSession.test.ts
+++ b/scripts/documentWatcherSession.test.ts
@@ -1,0 +1,12 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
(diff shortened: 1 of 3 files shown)


PathGao:
@alecdotdev This is the next dependent document-session PR after #276. It centralizes the self-write grace decision while retaining window listener wiring in MarkdownViewer.

alecdotdev (approved):

PR #278, opened 2026-07-30
refactor: move close decisions into document session

PathGao:
## Summary

Document close decisions now live with the document lifecycle that owns dirty state and saving. The viewer keeps presentation-only callbacks for the confirmation dialog and notifications, so the close behavior remains unchanged while future document work has one ownership boundary.

Depends on #277. Merge after it.

## Validation

- `npm run check`
- `npm test` (119 passed)
- `cargo test` (20 passed)


PathGao:
This is the next stacked refactor after #277. Its branch is directly based on #277; please merge it after #277. Validation: npm run check, npm test (119 passed), and cargo test (20 passed).

alecdotdev (approved):

PR #279, opened 2026-07-30
feat: rebuild tab movement across windows

PathGao:
## Summary

Tabs can again move directly between live windows without saving or losing unsaved content. The destination acknowledges a brokered transfer before the source removes its tab; a dead or busy destination leaves the source unchanged.

The tab context menu identifies each destination window, the window menu can merge other windows here, and Cmd/Ctrl+Shift+M carries the active tab through windows in creation order (or detaches it when no destination exists).

Rebuilds the functional scope of #225 without its stale merge history. Depends on #278; merge after it.

## Validation

- `npm run check`
- `npm test` (122 passed)
- `cargo test` (20 passed)


PathGao:
This rebuilds the functional scope of stale draft #225 on the new componentization chain. Its branch is directly based on #278; please merge it after #278. Validation: npm run check, npm test (122 passed), and cargo test (20 passed).

alecdotdev (approved):

Issue #230, opened 2026-07-16
YouTube embeds broken by YouTube's late-2025 Referer enforcement (error 153) — at least on macOS/Linux (tauri:// origin)


PathGao:
The YouTube embed feature (image/link syntax → iframe player; CSP already allowlists `youtube.com`/`youtube-nocookie.com` in `frame-src`) currently renders a silent 16:9 blank box on macOS. No error is surfaced; users cannot tell why.

## Root cause

Since late 2025 YouTube's embedded player refuses to initialize unless the request identifies its embedding origin via the `Referer` header ("Error 153: Video player configuration error"). This broke many regular websites too ([Simon Willison's writeup](https://simonwillison.net/2025/Dec/1/youtube-embed-153-error/)). Websites can fix it with `Referrer-Policy: strict-origin-when-cross-origin` — **but Markpad cannot**: on macOS/Linux the page origin is `tauri://localhost`, and browsers never send a Referer from non-http(s) origins. There is nothing to send.

Experiment matrix (macOS, same video, network confirmed fine):

| parent page | Referer sent | result |
|---|---|---|
| `http://127.0.0.1` test page | yes | ✅ plays |
| direct navigation to `/embed/…` | none | ❌ error 153 |
| `file://` test page | none (spec-forbidden) | ❌ error 153 |
| Markpad (`tauri://localhost`) | none (spec-forbidden) | ❌ blank box |

## Open question for maintainers: does Windows still work?

On Windows WebView2 the origin is `http://tauri.localhost` — a legit http origin that *does* send a Referer, so the feature may still be alive there. Worth a quick test before deciding scope.

## Options

1. **Degrade to thumbnail + open-externally** (suggested): render `https://img.youtube.com/vi/<id>/hqdefault.jpg` (already allowed by `img-src https:`, works without Referer) and open the video in the system browser on click. Predictable, immune to future policy changes; could be macOS/Linux-only if Windows embeds still work.
2. **Remove the feature**: links stay links. Also allows dropping `frame-src` entirely — the embed iframe is the only external-code execution point in the preview webview.
3. **`useHttpsScheme`** (Tauri v2) to make the origin `https://tauri.localhost` so a Referer exists. Heavy caveat: localStorage is partitioned by origin, so all existing users would silently lose settings/recent-files/session state unless a migration is shipped.

Happy to send a PR once maintainers pick a direction (and report Windows status).

alecdotdev:
[...]

PR #280, opened 2026-07-30
feat: rebuild window tags and pinned sessions

PathGao:
## Summary

Windows can be given a named color identity, which is reflected in the native title and cross-window destination list. Pinning an identity saves the window's file set at close; the HOME page can reopen that set and unpin it.

Rebuilds the functional scope of #226 without its stale merge history. Depends on #279; merge after it.

## Validation

- `npm run check`
- `npm test` (125 passed)
- `cargo test` (20 passed)


PathGao:
This rebuilds the functional scope of stale draft #226. Its branch is directly based on #279; please merge it after #279. Validation: npm run check, npm test (125 passed), and cargo test (20 passed).

PathGao:
Updated after the latest chain rebase: the Home menu now has a dedicated window-organization group immediately below export actions. It exposes `Window Tag…` (create or edit the current window tag) and `Merge All Windows Here`; a regression test locks their order. `npm run check`, `npm test` (126 tests), and `cargo test` (20 tests) pass locally.

PathGao:
@alecdotdev As the stack is almost through, could you also take a look at #230 and #281 when convenient? They capture two follow-up product decisions: the YouTube embed fallback/compatibility direction and the minimal macOS app-menu boundary.

PR #293, opened 2026-07-31
fix: preserve tabs after document read errors

### LINKED ISSUE #292: macOS: transient file-read errors close the current tab
## Problem

`loadMarkdown` closes the active tab whenever a read reports a missing file. A file watcher can observe a transient gap while another process atomically replaces the file, so this can remove a tab even though the document was just saved and remains recoverable.

## Expected behavior

Keep the tab and its in-memory buffer on every load failure. Surface the error and let the user decide whether to retry, save, or close it.

## Scope

Do not remove the recent-file entry or close the tab from the load error handler.

PathGao:
Fixes #292

Keep the tab, its in-memory buffer, and the recent-file entry when a document read fails. A transient missing-file event during an external atomic replacement can no longer close a recoverable document.

Depends on #288. Please merge in this order: #286 → #287 → #288 → this PR.

Validation:
- npm ci
- npm run check
- npm test
- cargo test

### DIFF
--- a/scripts/documentLoadFailure.test.ts
+++ b/scripts/documentLoadFailure.test.ts
@@ -1,0 +1,10 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const session = readFileSync('src/lib/sessions/documentSession.svelte.ts', 'utf8');
+
+test('a transient missing-file read error preserves the open tab', () => {
+	assert.doesNotMatch(session, /tabManager\.closeTab/);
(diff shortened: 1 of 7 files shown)


PathGao:
@alecdotdev This addresses the tab-closing path observed around transient file replacement. It is ready after #286.

PathGao:
@alecdotdev I rebased the pending reliability line onto the existing PDF/window-capability chain to eliminate the two parallel branches. The single merge order is now #286 → #287 → #288 → #293 → #296 → #298; the complete rebased chain passes npm check, 132 Node tests, and 20 Rust tests.

alecdotdev (approved):

PR #294, opened 2026-07-31
ci: sign and notarize macOS releases

### LINKED ISSUE #209: macOS: file-access permission not persisted on save (release is ad-hoc signed / not notarized)
**Describe the bug**

On macOS, Markpad repeatedly asks for file-access permission every time it **saves** a file into a protected folder (Desktop / Documents / Downloads), even after the user has explicitly allowed access in System Settings and even after granting **Full Disk Access** to the app. The grant never "sticks" — the prompt returns on the next save.

Opening files does **not** trigger the prompt (the macOS Open panel / Powerbox implicitly grants access to the user-selected file). The problem is specific to the write/save path.

**Likely root cause**

The distributed macOS app is **ad-hoc signed and not notarized**, so macOS cannot pin a TCC / Full Disk Access grant to a stable code-signing identity, and re-prompts on each protected-folder write. Evidence from the installed app (`/Applications/Markpad.app`):

```
$ spctl -a -vvv /Applications/Markpad.app
/Applications/Markpad.app: rejected
source=no usable signature

$ codesign -dvv /Applications/Markpad.app
Identifier=Markpad-60fe2234e2ef4c98
Signature=adhoc
TeamIdentifier=not set

# no sandbox / files entitlements
$ codesign -d --entitlements :- /Applications/Markpad.app   # (empty)
```

Because the signature is ad-hoc (no Developer ID, unstable auto-generated identifier), TCC treats the grant as untrusted / non-persistent. Granting Full Disk Access does not resolve it.

This was verified on the **official v2.6.11 release** (downloaded via browser, not a local build) — the app still carries its download quarantine attribute:

```
$ xattr -l /Applications/Markpad.app
com.apple.quarantine: 03c1;........;Chrome;........
# CFBundleShortVersionString = 2.6.11, binary dated Jun 2026
```

**To Reproduce**
1. On macOS, open a Markdown file located on the Desktop (or in Documents/Downloads).
2. Edit it and save (⌘S) back to the same location.
3. macOS shows a "Markpad wants to access files in your Desktop folder" prompt. Click Allow.
4. Grant Markpad **Full Disk Access** in System Settings → Privacy & Security.
5. Edit and save again — the permission prompt still appears. The grant is not remembered.

**Expected behavior**

After allowing access once (or after granting Full Disk Access), saving to Desktop / Documents / Downloads should not prompt again. The grant should persist across saves and app launches.

**Suggested fix**

Sign the macOS release with a **Developer ID Application** certificate and **notarize** it (staple the ticket). TCC / Full Disk Access grants are keyed to a stable code-signing identity; a Developer-ID signature gives macOS a stable identity to persist the grant against. An ad-hoc signature does not.

This was confirmed on the official downloaded v2.6.11 release (not a local build), which `spctl` reports as having `no usable signature`.

**Device**
 - OS: macOS (Darwin 25.5.0)
 - App Version: v2.6.11 (official release, downloaded)

**Additional context**

The app is not sandboxed, so file access to Desktop/Documents/Downloads is governed by TCC folder permissions rather than security-scoped bookmarks. Saving is a direct filesystem write to the target path, which is what triggers the TCC check on every write when the signing identity is not trusted/persistent.

PathGao:
Fixes #209

The release workflow now imports a Developer ID Application certificate, passes the Apple notarization credentials to the macOS build, and fails instead of publishing another ad-hoc-signed macOS bundle when the identity is unavailable.

The release runbook documents the six required repository secrets. A maintainer must configure them before the next release.

Depends on #293, then #286, because this branch includes their CI-baseline fixes; merge in that order.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -228,6 +228,30 @@ jobs:
           fi
 
       # --- MacOS Build ---
+      - name: Import Apple Developer ID certificate
+        if: matrix.platform == 'macos-latest'
+        shell: bash
+        env:
+          APPLE_CERTIFICATE: ${{ secrets.APPLE_CERTIFICATE }}
+          APPLE_CERTIFICATE_PASSWORD: ${{ secrets.APPLE_CERTIFICATE_PASSWORD }}
+          KEYCHAIN_PASSWORD: ${{ secrets.KEYCHAIN_PASSWORD }}
+        run: |
+          set -euo pipefail
+          echo -n "$APPLE_CERTIFICATE" | base64 --decode > certificate.p12
+          security create-keychain -p "$KEYCHAIN_PASSWORD" build.keychain
+          security default-keychain -s build.keychain
+          security unlock-keychain -p "$KEYCHAIN_PASSWORD" build.keychain
+          security set-keychain-settings -lut 21600 build.keychain
+          security import certificate.p12 -k build.keychain -P "$APPLE_CERTIFICATE_PASSWORD" -T /usr/bin/codesign
+          security set-key-partition-list -S apple-tool:,apple:,codesign: -s -k "$KEYCHAIN_PASSWORD" build.keychain
+          identity=$(security find-identity -v -p codesigning build.keychain | sed -n 's/.*"\(Developer ID Application:.*\)"/\1/p' | head -1)
+          if [ -z "$identity" ]; then
+            echo 'No Developer ID Application signing identity was imported.' >&2
+            exit 1
+          fi
+          echo "APPLE_SIGNING_IDENTITY=$identity" >> "$GITHUB_ENV"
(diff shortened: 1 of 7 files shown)


PathGao:
@alecdotdev This is ready for review after #293 and #286. Before the next release, please add the six Apple secrets documented in the PR; without them the new macOS release gate will fail rather than publish another ad-hoc-signed bundle.

PathGao:
Closing this PR in favor of #209. The workflow is correct only after an Apple Developer Program owner can provide a Developer ID Application certificate and notarization credentials; without them it would merely make every release fail. This is a release-ownership and credential prerequisite, not an application-code fix.

PathGao:
Tracked in #209; see the issue for the required Apple-side setup and the interim workaround.

PR #296, opened 2026-07-31
fix: route Live Mode changes to the watched file

### LINKED ISSUE #295: Live Mode can reload the wrong tab after an external file change
## Problem

Live Mode currently watches one file per window, but the watcher emits a pathless `file-changed` event. The frontend then reloads whichever tab is active at delivery time rather than the file that was watched.

This can overwrite unsaved edits in the active tab:

1. Open `A.md`, then `B.md`.
2. Enable Live Mode while `B.md` is active, so `B.md` is watched.
3. Switch to `A.md` and make an unsaved edit.
4. Modify `B.md` outside Markpad.
5. The `B.md` watcher event reloads `A.md`; its in-memory buffer is replaced and its dirty state is cleared.

## Root cause

`src-tauri/src/window_runtime.rs` emits `file-changed` with `()` and `src/lib/MarkdownViewer.svelte` reloads `currentFile`, which is derived from the active tab. The event has no way to identify the watched path.

## Expected behavior

A file-change event must carry the watched path. The UI must reload only when that path is the active Live Mode document; an event for another tab must never replace the active tab's buffer.

## Scope

Keep the single-watcher model. Route the watched path through the event and add a regression test for the cross-tab case.

PathGao:
Fixes #295

Live Mode now tracks the active file at the window layer and carries the watched path in file-change events. An event for a background tab can no longer reload the active tab or discard its unsaved buffer. Enabling Live Mode no longer reloads the current document.

Depends on #293. Please merge in this order: #286 → #287 → #288 → #293 → this PR.

Validation:
- npm ci
- npm run check
- npm test
- cargo test

### DIFF
--- a/scripts/documentLoadFailure.test.ts
+++ b/scripts/documentLoadFailure.test.ts
@@ -1,0 +1,10 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const session = readFileSync('src/lib/sessions/documentSession.svelte.ts', 'utf8');
+
+test('a transient missing-file read error preserves the open tab', () => {
+	assert.doesNotMatch(session, /tabManager\.closeTab/);
+	assert.match(session, /options\.onError\('Error loading file', error\);/);
+});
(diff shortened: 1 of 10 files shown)


PathGao:
@alecdotdev This is the next dependent safety fix after #293. It prevents Live Mode from reloading a different, dirty tab when a background watched file changes. All required checks pass; merge order is #286 → #293 → #296.

PathGao:
The branch is now directly based on #293; please use the updated single-chain merge order in the PR description.

alecdotdev (approved):

PR #298, opened 2026-07-31
fix: allow opening multiple documents

PathGao:
Fixes #297

The Open File dialog now accepts multiple document selections and loads each selected path through the existing tab-management flow.

Depends on #296. Please merge in this order: #286 → #287 → #288 → #293 → #296 → this PR.

Validation:
- npm ci
- npm run check
- npm test (132 passing on the complete rebased chain)
- cargo test (20 passing on the complete rebased chain)

Fixes #124.

PR #300, opened 2026-07-31
fix: retain all macOS open-document paths

### LINKED ISSUE #299: macOS open-document events ignore every path after the first
## Problem

The macOS `RunEvent::Opened` handler reads only `urls.first()`, and startup loading consumes only `args[0]`. Opening several Markdown documents in one Finder/open invocation therefore discards every path after the first even in a single, already-running bundle.

This is distinct from #212's second-copy/single-instance Apple Event loss: it fixes the direct multi-document path while that upstream timing limitation remains.

## Expected behavior

Every file URL delivered in a macOS open-document event reaches the receiving viewer and is opened through the normal tab flow, preserving the delivered order.

## Scope

Store and drain all startup paths, emit all delivered paths to an existing viewer, and load all startup arguments. Add a regression test for the Rust-to-frontend path collection contract.

PathGao:
Fixes #299

macOS `RunEvent::Opened` handling now retains every delivered file URL, drains all startup paths in delivery order, and opens every startup document through the normal tab flow. This is a direct multi-document fix; #212 remains open for the distinct second-copy single-instance Apple Event timing problem.

Depends on #298. Please merge in this order: #286 → #287 → #288 → #293 → #296 → #298 → this PR.

Validation:
- npm run check
- npm test (133 passing)
- cargo test (20 passing)
- macOS runtime Apple Event smoke test remains required after a build.

### DIFF
--- a/scripts/documentLoadFailure.test.ts
+++ b/scripts/documentLoadFailure.test.ts
@@ -1,0 +1,10 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const session = readFileSync('src/lib/sessions/documentSession.svelte.ts', 'utf8');
+
+test('a transient missing-file read error preserves the open tab', () => {
+	assert.doesNotMatch(session, /tabManager\.closeTab/);
+	assert.match(session, /options\.onError\('Error loading file', error\);/);
+});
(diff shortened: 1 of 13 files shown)


PathGao:
@alecdotdev This extends the same chain with a scoped macOS open-document fix. It deliberately does not claim to resolve #212’s second-bundle/single-instance timing limitation. All local checks pass; please merge after #298.

PR #302, opened 2026-07-31
fix: stop associating plain text files

PathGao:
Fixes #179

Release metadata no longer advertises `.txt` as a Markpad file association. This prevents installers from replacing the normal Text Document context-menu association while retaining `.md` and `.markdown` support.

Depends on #301. Please merge in this order: #286 → #287 → #288 → #293 → #296 → #298 → #300 → #301 → this PR.

Validation:
- Regression test verifies only Markdown extensions are advertised.
- npm run check
- npm test (134 passing)
- cargo test (21 passing)
- Windows installer smoke test remains required for a released package.

PR #301, opened 2026-07-31
fix: normalize imported VS Code theme slugs

PathGao:
Fixes #187

VS Code theme labels now normalize punctuation runs into one separator before matching the vscodethemes.com URL slug. This makes the documented SynthWave `'84` label match the `synthwave-84` URL segment.

Depends on #300. Please merge in this order: #286 → #287 → #288 → #293 → #296 → #298 → #300 → this PR.

Validation:
- Reproduced the mismatch against the upstream SynthWave package manifest (`SynthWave '84` versus `synthwave-84`).
- npm run check
- npm test (133 passing)
- cargo test (21 passing)

PR #304, opened 2026-07-31
Fix split scroll sync triggered by typing

### LINKED ISSUE #303: Split scroll sync repositions panes on every cursor move while typing
## Problem

When split scroll sync is enabled, the editor registers `onDidChangeCursorPosition` as a scroll-sync source. Typing moves the cursor on every character, so each edit emits a new scroll position. If preview layout changes while rendering, the reciprocal positioning makes both panes visibly jump.

This is the remaining typing-jitter symptom reported in #205; it is distinct from normal scroll synchronization.

## Expected behavior

Only actual editor scroll changes should initiate editor-to-preview synchronization. Cursor movement and text insertion without scrolling must not reposition either pane.

PathGao:
Fixes #303

This removes the cursor-position listener from editor-to-preview scroll synchronization. A normal typing edit can move the cursor without representing an editor scroll, so it must not reposition the preview pane. Real editor scroll events continue to synchronize as before.

Regression coverage verifies that the synchronization effect listens to `onDidScrollChange` and not `onDidChangeCursorPosition`.

Depends on #302. Merge order: #286 → #287 → #288 → #293 → #296 → #298 → #300 → #301 → #302 → this PR.

Validated with `npm ci`, `npm run check`, `npm test` (135 tests), and `cargo test` (21 Rust tests).

### DIFF
--- a/scripts/documentLoadFailure.test.ts
+++ b/scripts/documentLoadFailure.test.ts
@@ -1,0 +1,10 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const session = readFileSync('src/lib/sessions/documentSession.svelte.ts', 'utf8');
+
+test('a transient missing-file read error preserves the open tab', () => {
+	assert.doesNotMatch(session, /tabManager\.closeTab/);
+	assert.match(session, /options\.onError\('Error loading file', error\);/);
+});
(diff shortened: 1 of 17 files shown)


PathGao:
@alecdotdev This is the next focused reliability fix in the existing chain. It removes the cursor-change trigger that made split panes jump during normal typing; it retains only actual editor-scroll synchronization.

Please merge after #302, following the chain stated in the PR description.

PR #310, opened 2026-07-31
Update vulnerable Svelte build toolchain

PathGao:
Fixes #309

Updates SvelteKit, Svelte, Vite, and the matching adapter/plugin/check packages within their existing major release lines. The lockfile refresh also updates affected transitive packages.

Pins `cookie` to 0.7.2 because the latest SvelteKit 2 dependency range still resolves its vulnerable 0.6.0 release.

The complete resolved lockfile is clean under `npm audit`.

This is the next stacked change after #308; merge the already-listed predecessor PRs first, then this PR.

Validated with `npm ci`, `npm run check` (0 errors; one Settings drag-header accessibility warning detected by the newer checker is being handled separately), `npm test` (136 passing), `npm run build`, `npm audit` (0 vulnerabilities), and `cargo test` (21 passing).

PR #312, opened 2026-07-31
Keep Settings modal drag accessible

### LINKED ISSUE #311: Restore an accessibility-clean Settings modal after Svelte checker upgrade
After the Svelte toolchain update in #310, `npm run check` reports an accessibility warning for the pointer-drag handler attached directly to `.settings-header`. The header contains the close button, so assigning an interactive role to the header would create invalid nested controls.

Keep the existing pointer drag behavior while moving event ownership to the already-semantic dialog and explicitly limit drag starts to the non-interactive header surface. Add a regression test and require `npm run check` to return with no warnings.

PathGao:
Fixes #311

Moves the pointer listener from the non-interactive header to the existing dialog element, while starting a drag only when the event originated from the header’s non-interactive surface. The close button remains excluded. This preserves mouse dragging without giving a header that contains a button an invalid interactive role.

Adds a regression test for the event ownership and scope.

This is stacked after #310; merge the predecessor chain first.

Validated with `npm run check` (0 errors, 0 warnings), `npm test` (136 passing), `npm run build`, and `cargo test` (21 passing).

### DIFF
--- a/scripts/toolbarCustomizationWiring.test.ts
+++ b/scripts/toolbarCustomizationWiring.test.ts
@@ -63,9 +63,13 @@ test('settings modal uses a dedicated resize handle instead of native CSS resize
 	assert.match(settingsComponent, /\.settings-modal[\s\S]*max-height:/);
 });
 
-test('settings modal can be dragged by the header and resized from every edge', () => {
+test('settings modal starts dragging only from the non-interactive header surface', () => {
 	assert.match(settingsComponent, /handleSettingsModalDragPointerDown/);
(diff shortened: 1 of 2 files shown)


PathGao:
@alecdotdev Follow-up to #310: the upgraded checker exposed one Settings drag-header accessibility warning. This keeps the existing pointer drag behavior and restores a clean check with a focused regression test. Please review after #310 and its predecessors.

alecdotdev (approved):

PR #314, opened 2026-07-31
Audit complete dependency lockfile in CI

PathGao:
Fixes #313

Changes the pull-request security gate from a production-only audit to `npm audit` of the complete resolved lockfile. Build dependencies parse project input during development and release builds, so excluding them hid the SvelteKit/Svelte/Vite findings addressed in #310.

Updates the workflow regression test to require the complete audit and reject a production-only command.

This is stacked after #312 (and therefore #310). Merge the predecessor chain first.

Validated with `npm ci`, `npm audit` (0 vulnerabilities), `npm run check` (0 errors, 0 warnings), `npm test` (136 passing), `npm run build`, and `cargo test` (21 passing).

Issue #205, opened 2026-07-10
Scroll Sync Do Not Function


defencedog:
**Describe the bug**
Scroll Sync icon doesn't function when side by side prevview windows is open

**To Reproduce**
Whether you enable or disable scroll sync icon there is no effect

**Expected behavior**
When I scroll both source code & preview should scroll in sync

**Device**
 - OS: Win11 x64
 - App Version v2.6.11

PathGao:
Scroll sync was rewritten in #196 (pixel-based sync positions instead of line/ratio), which first shipped in **v2.6.12** — released 2026-07-13, after this report was filed against v2.6.11. Could you retest on v2.6.12? If sync works there, this issue can be closed.

Guardiancelte:
looks like it is working. but sync is not perfect, I cannot share the exact file that I was using so I ask GPT to to give me a test file. 

When reaching the end of the file it is quite a few lines off. I had this issue in both stress test (I asked a easier one from GPT as I thought it might be the problem with too many format and tables and charts in the hard one, but actually was in both)

[Stress Test Easy.md](https://github.com/user-attachments/files/30209526/Stress.Test.Easy.md)
[Stress Test Hard.md](https://github.com/user-attachments/files/30209525/Stress.Test.Hard.md)

<img width="2826" height="1901" alt="Image" src="https://github.com/user-attachments/assets/628354ad-1ca4-4b96-a0c9-fca2dba60502" />

Another issue which shows a lot more on the hard one, but a little on the easy one as well, is when typing text on left side, both the right and left panel are jumping up and down for every character typed (not even doing return to the line, just typing normal chararcters). Which is highly disturbing. 

Otherwise I am actually impressed by the displaying capabilities.

PathGao:
I traced the typing-related jump to `Editor.svelte`: editor-to-preview sync subscribed to `onDidChangeCursorPosition`, so every ordinary edit repositioned the paired pane. #304 removes that trigger and adds a regression test; real editor scroll events remain synchronized.

The reported end-of-document drift is a separate mapping/measurement problem and stays tracked here for a reproducible fixture and platform retest.

PathGao:
I traced the remaining end-of-document offset to Monaco’s `scrollBeyondLastLine` padding being included in the pixel-ratio sync range. #316 calculates the editor sync range from content height instead, while preserving the editor’s padding preference and the front-matter split. It follows #304; please retest the attached fixtures after the stack merges.

[...]

PR #316, opened 2026-07-31
Exclude editor bottom padding from split scroll sync

### LINKED ISSUE #315: Exclude Monaco bottom padding from split-scroll synchronization
The end-of-document drift still reported in #205 has a concrete cause in the current pixel-ratio mapping: `Editor.svelte` enables Monaco `scrollBeyondLastLine`, then uses `getScrollHeight() - layout.height` as the synchronization range. The extra bottom padding is not document content, so reaching the final rendered line produces a ratio below 1 and leaves the preview visibly above its end. The inverse direction can place the editor in blank padding.

Keep the editor preference, but calculate the editor synchronization range from Monaco content height rather than scroll height. Use the same content range for both emitting and applying sync positions; retain the existing front-matter segmentation. Add a regression test for the invariant. This follows #205 and #304.

PathGao:
Fixes #315

Monaco keeps blank space below the final line because `scrollBeyondLastLine` is enabled. The previous synchronization range used scroll height, so that non-content space made the editor’s final document line map to a preview position above its end.

Uses Monaco content height for both directions of split-scroll synchronization while retaining the editor’s scroll-beyond-last-line preference and front-matter segmentation.

Adds a regression assertion that synchronization does not use the padded scroll range.

This is stacked after #314 and its predecessors. Merge the predecessor chain first.

Validated with `npm run check` (0 errors, 0 warnings), `npm test` (136 passing), `npm run build`, and `cargo test` (21 passing).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -27,8 +27,8 @@ jobs:
       - name: install frontend dependencies
         run: npm ci
 
-      - name: audit production dependencies
-        run: npm audit --omit=dev
+      - name: audit complete dependency lockfile
+        run: npm audit
 
       - name: check frontend
         run: npm run check
--- a/package-lock.json
+++ b/package-lock.json
@@ -27,15 +27,15 @@
 				"yaml": "^2.9.0"
 			},
 			"devDependencies": {
-				"@sveltejs/adapter-static": "^3.0.6",
-				"@sveltejs/kit": "^2.9.0",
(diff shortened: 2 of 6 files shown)


PathGao:
@alecdotdev This is the follow-up for the remaining end-of-document drift reported in #205. It keeps Monaco’s normal bottom-padding preference but removes that padding from the sync calculation. It is stacked after #314 and the predecessors; please review in order.

PR #318, opened 2026-07-31
Reuse the created window handle during startup

### LINKED ISSUE #317: Avoid a startup panic if a just-created window is no longer in the registry
The startup setup closure builds a Tauri window successfully, then immediately looks it up twice by label with `app.get_webview_window(label).unwrap()`. The newly returned window handle already supplies the same operations. If the registry changes during startup/teardown, the lookup can turn a recoverable lifecycle race into a process panic.

Use the handle returned by `WebviewWindowBuilder::build()` for theme, file delivery, focus, installer sizing, and centering. Add a regression test that rejects registry lookup unwraps in that setup path.

PathGao:
Fixes #317

Uses the handle returned by `WebviewWindowBuilder::build()` for initial theme application, file delivery, focusing, installer sizing, and centering. This removes two registry lookup `unwrap`s that could panic if window lifecycle state changes during startup.

Adds a regression test that rejects those startup lookup unwraps.

This is stacked after #316 and its predecessors. Merge the predecessor chain first.

Validated with `npm ci`, `npm audit` (0 vulnerabilities), `npm run check` (0 errors, 0 warnings), `npm test` (137 passing), `npm run build`, and `cargo test` (21 passing).

### DIFF
--- a/scripts/windowStartupHandle.test.ts
+++ b/scripts/windowStartupHandle.test.ts
@@ -1,0 +1,10 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+const backend = readFileSync('src-tauri/src/lib.rs', 'utf8');
+
+test('startup reuses the window handle returned by the builder', () => {
+	assert.match(backend, /let window = window_builder\.build\(\)\?;/);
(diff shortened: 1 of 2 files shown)


PathGao:
@alecdotdev This startup-lifecycle guard removes two avoidable `unwrap`s after window creation. It is stacked after #316 and its predecessors; please review in that order.

PR #325, opened 2026-07-31
fix(preview): migrate full-width preference

PathGao:
## Summary

- Moves Full width persistence from the generic `isFullWidth` key to `preview.fullWidth`.
- Preserves existing users’ setting when the new key is absent; a valid new-key value wins when both exist.
- Removes the legacy key after the current state is persisted.

## Validation

- `npm run check`
- `npm test` (149 passing)
- `npm run build`
- `cargo test` (25 passing)

## Stack

Closes #263. This PR follows #320 → #321 → #322 → #323 → #324.

## Post-deploy monitoring

- Verify current Full width users retain their selection across the first upgraded launch.
- Verify a changed Full width selection persists to the dedicated key.

PR #326, opened 2026-07-31
fix(math): preserve display math subscripts

### LINKED ISSUE #174: math rendering problem with \bar, \vec and possibly others
**Describe the bug**
When using math equations with \bar or \vec with subscripts, the equations are not rendered correctly in some situations.

**To Reproduce**
Type at least two variables with \bar or \vec and subscripts that have letters embraced by { }, and the rendering will fail.
In order to render correctly, only the first one could use subscript with { }.
If a space is used in front of the first subscript with { } it seems to render correctly.

**Expected behavior**
I expect all the formulas to be rendered correctly.

**Screenshots**
An array of examples of what works and what doesn't is shown in the following picture:

Image

**Device**
 - OS: Windows 11
 - Version 25H2
 - App Version 2.6.10

PathGao:
## Summary

Display equations with multiple subscripts now render as equations instead of showing literal `$$…$$` text. Comrak was applying Markdown emphasis before KaTeX received the display-math source; the backend preserves underscores within display delimiters and restores them after conversion.

Fixes #174.

## Validation

- `npm run check`
- `npm test` (149 passing)
- `cargo test` (27 passing)
- Built and opened an isolated macOS `Markpad PR Test.app` (separate bundle identifier); all five formulas supplied in #174 rendered correctly.

## Merge order

This is the next commit in the current chain and should merge after #325.

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -365,6 +365,24 @@ mod tests {
     fn theme_slug_collapses_punctuation_runs() {
         assert_eq!(theme_slug("SynthWave '84"), "synthwave-84");
     }
+
+    #[test]
+    fn display_math_keeps_multiple_braced_subscripts_out_of_markdown_emphasis() {
+        let html = convert_markdown("$$\\bar{b}_{1} + \\bar{b}_{2}$$\n");
+        assert!(
+            html.contains("$$\\bar{b}_{1} + \\bar{b}_{2}$$"),
+            "unexpected parser output: {html}",
+        );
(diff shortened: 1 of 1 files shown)


PathGao:
@alecdotdev #326 fixes the newly reproduced #174 regression. It is the next commit after #325; the attached input was verified in an isolated macOS release bundle, not `/Applications/Markpad.app`.

PathGao:
Superseded by native Stack PR #351 (Stack #353). Closing this fork-head PR so #351 is the active review path.

Issue #232, opened 2026-07-21
Export to PDF Format is Broken


Guardiancelte:
**Describe the bug**
Exporting to PDF, format is broken 

**To Reproduce**
1. Open a .md
2. Click export to PDF
3. Preview shows a lot of overlaps and only aligned to the left

**Expected behavior**
I would expect the same format as HTML, which right now I convert to HTML and then do print using my browser which then the pdf format has no issues

**Screenshots**
Hard stress test
< img width="2291" height="2043" alt="Image" src="https://github.com/user-attachments/assets/cd30890d-3a39-4991-80cd-aada30e8ce2c" />

Easy stress test
< img width="2209" height="2052" alt="Image" src="https://github.com/user-attachments/assets/58ed21f5-00fe-45ec-a190-26c5a5c89622" />

**Device**
 - OS: Windows 11
 - App Version: 2.6.12

**Additional context**
I created 2 test files with GPT as I could share my private data, a easy and hard one. The easy one I would expect to be handled without issues to be honest.

[Stress Test Easy.md](https://github.com/user-attachments/files/30209738/Stress.Test.Easy.md)
[Stress Test Hard.md](https://github.com/user-attachments/files/30209737/Stress.Test.Hard.md)

PathGao:
@Guardiancelte This should be addressed by [#266](https://github.com/alecdotdev/Markpad/pull/266), merged on July 29. It removes the stale fixed heights from folded sections during PDF printing, allowing the reflowed content to occupy its correct height instead of overlapping. Could you please try the latest build and close this issue if the PDF output is now correct?

pcuttriss:
@PathGao  Just tried a pdf export on a file that was not formatting correctly a couple of day ago and it looks good, I also took a look at the stress docs linked above and they look ok to me......
Hopefully @Guardiancelte will have similar feedback.

Guardiancelte:
Will have a look next week.

Guardiancelte:
ok, I retested with my easy stress test file, there is an improvement when split view is not toggled the pdf is now correct but when split view is toggled (there is less messiness in the pdf with no title overlap so that is good) it is still formatted as if the text was restricted to a left column? I am assuming that is not desired behaviour.

side note, do you know why the pdf is marked as tauri.localhost? I feel like that is a bit useless, is that a leftover of of the tool used to convert to pdf or something? IMO should be removed or have a option to have a author field

not toggled

[...]

PR #327, opened 2026-07-31
fix(preview): persist task checkbox changes

PathGao:
## Summary

Preview checklist changes now follow the checkbox's committed state, so mouse, keyboard, and accessibility input all write the intended Markdown marker. The former click handler inferred a new state and could leave the file unchanged when the input provider had already updated the native control.

Fixes #148.

## Validation

- `npm run check`
- `npm test`
- `cargo test`
- Built an isolated macOS `Markpad PR Test.app`; clicking a preview checkbox changed it from unchecked to checked and wrote `- [x]` to disk.

## Merge order

This is the next commit in the current chain and should merge after #326.

PR #329, opened 2026-07-31
fix(print): use native controls and full page width

PathGao:
## Summary

- restore the macOS/Linux native print sheet through Tauri `WebviewWindow::print()`; the browser-level `window.print()` path can silently do nothing in the bundled app
- export directly to a user-chosen file through WebView2 on Windows, with print headers and footers disabled so PDFs do not contain `tauri.localhost`
- retain the full-width print layout when exporting from split view
- surface PDF-export failures as an in-app toast and bound the Windows completion wait to 60 seconds

## Platform behavior

- **macOS/Linux:** `Export as PDF` opens the native print sheet, including the platform's existing “Save as PDF…” flow.
- **Windows:** `Export as PDF` opens a PDF save dialog, then uses WebView2 `PrintToPdf` with headers/footers disabled.

## Validation

- `npm ci`
- `npm run check` — 0 errors, 0 warnings
- `npm test` — 128 passing
- `cargo test`
- `cargo check --target x86_64-pc-windows-msvc`
- built an isolated macOS release and exported the two reporter-provided stress documents from split view: the easy document was 29 A4 pages and the hard document was 92 A4 pages; inspected first, continuation, and table pages for full-width content and no URL footer.

Windows WebView2 output is compile-validated here; visual verification still needs a Windows runtime.

Related: #232

Depends on #337 until its test-only commit is merged.

### DIFF
--- a/scripts/issue261EditorPdf.test.ts
+++ b/scripts/issue261EditorPdf.test.ts
@@ -25,3 +25,10 @@ test('print layout releases measured fold heights before text reflows', () => {
 		/@media print\s*\{[\s\S]*?\.foldable-content-wrapper\.is-collapsed\s*\{[\s\S]*?height:\s*0\s*!important;/,
 	);
 });
+
+test('print layout lets the viewer pane escape the interactive split flex ratio', () => {
+	assert.match(
+		styles,
+		/@media print\s*\{[\s\S]*?\.pane\.viewer-pane\s*\{[\s\S]*?flex:\s*none\s*!important;/,
+	);
+});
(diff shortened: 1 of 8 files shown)


PathGao:
@alecdotdev This is the next linear follow-up after #327. It fixes a confirmed macOS PDF regression specific to exporting while split view is active: the viewer pane retained its inline split flex ratio, so the PDF body used only part of the printable page. I verified the minimal print-only change in an isolated release bundle with a 20-section A4 export. Please merge it after #327; I left it as a draft until that dependency lands.

PR #342, opened 2026-08-02
fix: open YouTube embeds externally

### LINKED ISSUE #230: YouTube embeds broken by YouTube's late-2025 Referer enforcement (error 153) — at least on macOS/Linux (tauri:// origin)

PathGao:
## Summary

- replace in-app YouTube iframes with clickable thumbnail cards
- open the original YouTube URL through Markpad's existing system-browser link handler
- prevent linked thumbnails from opening the preview image-zoom overlay
- remove the no-longer-needed YouTube frame CSP permission

Fixes #230

## Validation

- `npm ci`
- `npm run check`
- `npm test`
- `cargo test`

## Local bundle validation

- macOS Bundle: `com.alecdotdev.markpad.pr230`, version `2.6.14-pr230.local.1`
- Built with `npm run tauri build` in an isolated clone.
- Opened a Markdown document containing a YouTube URL; the thumbnail card rendered, and clicking it opened the original video in the system browser without opening Markpad's image-zoom overlay.


PathGao:
Superseded: #338 was already merged. The planning-document cleanup is consolidated in the dedicated follow-up PR.

Issue #334, opened 2026-07-31
SECURITY: the windows installer of the new version has a trojan virus


ajoe:
**Describe the bug**
Update to the latest version.

**To Reproduce**
1. Go to Setting
2. Click on Update

**Expected behavior**
Update without a virus.

**Screenshots**
<img width="1022" height="611" alt="Image" src="https://github.com/user-attachments/assets/17d82fa1-7fe5-442e-a379-8c0486b9ecb9" />

**Device**
 - OS: windows
 - Version 11
 - App Version 2.6.14

alecdotdev:
I'll look into this, seems like it could be a false positive.
I'll remove 2.6.14 until I addressed

alecdotdev:
Looks like it's a false positive caused by the custom uninstall script. Will rework it to not result in the warning. Thank you

PR #355, opened 2026-08-02
build: harden release inputs and isolate test bundles

PathGao:
## Summary

- Use `npm ci` in release builds so the resolved dependency graph matches the lockfile.
- Stop uploading portable Windows executables under installer-like names.
- Add an isolated macOS test-bundle builder requiring a non-production bundle identifier.
- Add workflow regression tests and update release documentation.

## Scope

This is independent of the current Stack. It is related to the release hardening tracked in #334, but does not claim to resolve the pending Defender investigation.

## Validation

- `npm ci`
- `npm run check`
- `npm test` (152 passing)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -110,7 +110,7 @@ jobs:
           sudo snap install snapcraft --classic
 
       - name: Install Frontend Dependencies
-        run: npm install
+        run: npm ci
 
       # --- Windows Build (x64) ---
       - name: Build Windows x64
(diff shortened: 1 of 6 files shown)


PathGao:
@alecdotdev This PR also removes the release-workflow behavior that uploaded the portable executable a second time under the misleading `MarkpadInstaller_*` name. The actual NSIS installer remains the `*-setup.exe` artifact.

For #334: I saw that the issue was closed after identifying the custom uninstall script as the likely Defender trigger, but I cannot find a corresponding code change. Do you want to rework that uninstall path before the next Windows release, or was the issue closed because the release was removed and no further change is planned?

PR #367, opened 2026-08-02
fix(editor): scope image undo to the tab that owns the image

PathGao:
## Summary

Undoing an image insert deletes the copied file — the only place the editor destroys something outside the buffer. Two ways of deciding *which* file were wrong, and both deleted one the user still had on screen.

### An undo in one tab could delete another tab's image

The editor keeps a single component instance across tab switches, so the managed list mixes every tab's images while the undo handler only ever looked at the newest entry.

```
tab A   drop image        managedImages = [ img/pic.png ]
           │
           │  switch to tab B  (component is not rebuilt)
           ▼
tab B   type, then Ctrl+Z
           handler reads the newest entry — tab A's
           B's buffer does not contain "![alt](img/pic.png)"
              of course not, it is a different document
           → deletes C:/notes/img/pic.png
             which tab A is still displaying
```

Entries now carry the tab that inserted them, and an undo only considers its own.

### Editing the caption read as removing the image

The check was `content.includes(embed)` against the whole embed string, so renaming `alt` to `logo` made the image look gone. The next undo — of anything — deleted a file the document was still pointing at, leaving a broken image the user never touched. The link target is what binds the buffer to the file, so that is what is matched now.

An embed that cannot be parsed falls back to the literal string: keeping a file that is no longer referenced is the cheaper mistake.

### Redo

Handled rather than ignored — the entry returns to the managed list, so a later undo cannot try to delete the same file twice.

**Known limitation, deliberately not fixed here:** the file itself stays deleted. Undo removes it immediately, and Ctrl+Z followed by Ctrl+Shift+Z therefore restores the embed text pointing at a file that is gone. Fixing that properly means deferring the deletion — for example reaping orphans when the tab closes — which is a behaviour change worth its own review rather than a rider on a data-loss fix.

## Also in this path

- The owning tab is captured **before** the first `await`. The copy and the edit are several round trips apart, so reading the active tab afterwards would misattribute an image if the user switched tabs while the copy was in flight.
- Closing a tab drops its bookkeeping, which otherwise grew for the lifetime of the window.

Follows #357, which fixed the directory and filename the deletion targets.

## Validation

- `npm run check` — 0 errors, 0 warnings
- `npm test` — 173/173. The decision logic is pure and directly executed by the tests: cross-tab isolation both ways, caption edits, look-alike filenames (`img/other-a.png` must not match `img/a.png`), the unparseable-embed fallback, redo re-adoption and cross-tab redo isolation, and closed-tab pruning.

### DIFF
--- a/scripts/managedImageUndo.test.ts
+++ b/scripts/managedImageUndo.test.ts
@@ -1,19 +1,136 @@
 import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
 import test from 'node:test';
-import { managedImageFromCopy } from '../src/lib/utils/managedImages.js';
+import {
+	contentReferencesImage,
+	embedTarget,
+	forgetClosedTabs,
+	imagePathOf,
+	managedImageFromCopy,
+	resolveManagedImageRedo,
+	resolveManagedImageUndo,
+	type ManagedImage,
+} from '../src/lib/utils/managedImages.js';
 
-test('undo metadata retains the collision-resolved image filename and paste-time directory', () => {
-	const image = managedImageFromCopy({
-		embed: '![alt](img/logo_1710000000.png)',
+const editor = readFileSync(new URL('../src/lib/components/Editor.svelte', import.meta.url), 'utf8');
+
+function image(tabId: string, relativePath: string, alt = 'alt'): ManagedImage {
+	return managedImageFromCopy({
+		tabId,
+		embed: `![${alt}](${relativePath})`,
 		parentDir: 'C:/notes',
 		imageDirectory: 'img',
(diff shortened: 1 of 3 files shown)


PathGao:
Closing this in favour of asking first.

This fixed the two ways the undo handler picked the wrong file, but it left the underlying design in place — and that design is the actual problem. Undo is the only one of six ways to remove an embed that touches the disk, and the deletion is an unlink rather than a move to the trash, so a wrong guess is unrecoverable.

That is a product decision rather than a bug fix, so it belongs with the maintainer. Opening an issue with the details; happy to implement whichever direction is preferred.

PR #417, opened 2026-08-03
docs(security): record what confines mermaid's generated stylesheet

PathGao:
No behaviour change. Comments and one tripwire test.

## Why this was worth measuring

`sanitizeDiagramSvg` lets mermaid's `<style>` through **on purpose** — the document policy's `FORBID_TAGS: ['style']` would strip it and leave every diagram an unstyled skeleton. That block is built partly from text the document controls, and it is injected into the app's live document. Whether a document can write CSS that escapes its own diagram had never been measured, only assumed.

**Audit:** ~110 payloads across 16 entry points — `classDef` (flowchart and the laxer class/state parsers), `style`, `linkStyle`, `themeVariables`, `fontFamily`, `themeCSS` via `%%{init}%%` **and** via YAML front matter, at-rules (`@import`/`@font-face`/`@media`/`@layer`/`@scope`/`@supports`/`@container`/`@keyframes`), `securityLevel` self-override, `<style>` inside labels and `accTitle`, textual `</style>` breakout — driven through the real Mermaid 11.16.0 / DOMPurify 3.4.12 / stylis 4.3.6 in a real browser, measuring `getComputedStyle` and `elementFromPoint` on probes placed **outside** the diagram. One run went over real HTTP with a logging server.

**Result: no payload reached a selector matching an ancestor of the diagram's `<svg>`.** `.titlebar{display:none}` and an exfiltrating `body::after` are both unreachable.

## What actually holds it up — and what does not

**Not `sanitizeDiagramSvg`.** It was measured to be a **byte-for-byte pass-through** for the style block in all 96 cases. It is the control everyone assumes is doing this job, and it never touches it.

The confinement is mermaid-internal, in four independent places:

1. **stylis' `addNamespace` prefixes every emitted selector with `#<svg-id>`, and CSS has no ancestor combinator** — a rule starting with that id can only ever match the SVG or its descendants, whatever follows. The id is one Markpad generates.
2. **`sanitizeCss`'s *running* brace-balance check** (depth may never go negative) is what stops a payload closing the `#<svg-id>{ … }` wrapper early. Overall balance would not be enough.
3. **The `CSSStyleSheet.replaceSync()` round-trip** re-serialises only rules the browser's parser accepted, so raw text — `</style>`, unbalanced constructs — cannot survive.
4. **`config.sanitize()`** deletes any string option containing `<`, `>` or `url(data:`, and any key on the `secure` list.

`securityLevel` is **not** part of this — it governs HTML inside labels, not CSS generation — and a document cannot lower it anyway (`secure` list).

## Two things held by accident, now pinned

**A. The `&` gap is real.** stylis substitutes `&` with the namespace, and `addNamespace` then *skips* a selector that already starts with it. So:

```
themeCSS: "& ~ *{display:none}"   →   #mermaid-…~*{display:none;}
```

— no prefix, and it was **measured live** to hide a `<span>` placed next to the `<svg>`. It reaches nothing in Markpad **only because the `<svg>` is the sole child of its container** everywhere it is mounted. That is a DOM-shape invariant, not a designed defence: a caption, a copy button, or an overlay added inside `.mermaid-diagram` re-arms it. Now documented at the container and pinned by a test that fails if anything is appended next to the svg.

**B. A document can cover the preview pane.** Making its own SVG `position:fixed` with `!important` produces an opaque full-pane overlay that does not scroll away. It cannot cross into the title bar only because `.markdown-body` carries `transform: translate3d(0,0,0)` — added for **scroll performance** — which makes it the containing block for fixed descendants. Someone optimising scrolling could remove that without knowing what else it holds, so it is pinned too.

## Correction to this file's own comment

`export.ts` claimed `classDef` is the route by which a hostile document lands a `url(https://…)` beacon in the block. **The flowchart lexer rejects `url()` in `classDef`, `style` and `linkStyle` outright.** The working route is `themeCSS`, set from `%%{init}%%` or YAML front matter — confirmed on the wire, 4/4 beacons fetched. The conclusion is unaffected: parity with a plain remote ``, which the export CSP allows precisely so author images keep working; **no new capability**.

## The test

`scripts/mermaidStyleScope.test.ts` — three source assertions plus a header carrying the audit's findings. These cannot prove browser behaviour; they are a tripwire that the two conditions the audit measured are still the ones in the tree. The audit is the evidence; this is what keeps it from silently expiring.

Scoped deliberately: the `appendChild` assertion covers only the block that builds and fills `.mermaid-diagram`, not the module — `.code-block-shell` legitimately appends, and forbidding it everywhere would be a rule about the wrong thing.

```
npm run check   435 files, 0 errors
npm test        503 / 503
```

## Coverage the audit did not achieve

- **Engine mismatch.** Everything ran in Chrome. Markpad ships on WKWebView (macOS), WebKitGTK (Linux) and WebView2 (Windows). Defence 3 depends on constructable stylesheets; `createCssStyles` calls `new CSSStyleSheet()` unconditionally, so an engine lacking it would fail the diagram outright rather than fall back — but that was not verified on WebKit. CSS parser strictness around `@scope`/`@container` also differs.
- **13 diagram types exercised, ~18 not.** `createUserStyles` is shared and per-diagram `getStyles` only receives regex-restricted `themeVariables`, so no difference is expected — **that is an inference, not a measurement**.
- **Not a proof of impossibility.** One gap was found in `addNamespace` and no other across ~110 payloads. "Tried hard, found only the `&` gap" is the claim; "no other gap exists" is not.

### DIFF
--- a/scripts/mermaidStyleScope.test.ts
+++ b/scripts/mermaidStyleScope.test.ts
@@ -1,0 +1,115 @@
+import assert from 'node:assert/strict';
+import { readFileSync } from 'node:fs';
+import test from 'node:test';
+
+/*
+ * Mermaid emits a `<style>` block built partly from text the document
+ * controls, and `sanitizeDiagramSvg` deliberately lets it through — the
+ * document policy's `FORBID_TAGS: ['style']` would strip it and leave every
+ * diagram an unstyled skeleton. That makes "can a document write CSS that
+ * escapes its own diagram" a real question, and it was audited rather than
+ * assumed: ~110 payloads across every entry point a document controls
+ * (`classDef`, `style`, `linkStyle`, `themeVariables`, `fontFamily`,
+ * `themeCSS` via `%%{init}%%` and via YAML front matter, at-rules, label HTML,
+ * `securityLevel` self-override, textual `</style>` breakout), driven through
+ * the real Mermaid 11.16.0 in headless Chrome, measuring `getComputedStyle`
+ * and `elementFromPoint` on probes placed OUTSIDE the diagram.
+ *
+ * Result: no payload reached a selector matching an ancestor of the diagram's
+ * `<svg>`, so `.titlebar{display:none}` and `body::after{background-image:…}`
+ * are both unreachable. What holds it up is Mermaid-internal, in four
+ * independent places — none of them `sanitizeDiagramSvg`, which was measured
+ * to be a byte-for-byte pass-through for the style block in all 96 cases:
+ *
+ *   1. stylis' `addNamespace` middleware prefixes every emitted selector with
+ *      `#<svg-id>`, and CSS has no ancestor combinator.
+ *   2. `sanitizeCss`'s RUNNING brace-balance check (depth may never go
+ *      negative) is what stops a payload closing the `#<svg-id>{ … }` wrapper
+ *      early. Overall balance would not be enough.
+ *   3. The `CSSStyleSheet.replaceSync()` round-trip re-serialises only rules
+ *      the browser's own parser accepted, so raw text — `</style>`, unbalanced
+ *      constructs — cannot survive into the block.
+ *   4. `config.sanitize()` deletes any string option containing `<`, `>` or
+ *      `url(data:`, and any key on the `secure` list (which includes
+ *      `securityLevel`, so a document cannot lower it).
+ *
+ * `securityLevel` is NOT part of this. It governs HTML inside diagram labels;
+ * it does not gate CSS generation.
+ *
+ * TWO THINGS THIS FILE PINS, because both are currently held by accident
+ * rather than by design:
+ *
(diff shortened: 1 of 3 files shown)


PathGao:
Closing this — it does not earn its place in the tree.

The audit found no vulnerability, so this PR ships no fix. What it ships is a 115-line test file whose assertions pin a DOM shape and the presence of a CSS declaration, plus a paragraph of security prose at a container. One of those assertions pins `.markdown-body`'s `translate3d` — a **scroll-performance** optimisation — as if it were a security control, so anyone legitimately changing scroll behaviour would get a confusing failure about mermaid.

The audit was worth running: it replaced an assumption with a measurement, and it corrected a claim in `export.ts` about which mermaid directive can carry a `url()`. But that value is ours, not the repository's, and it does not justify adding tests that constrain future refactoring against a threat that does not exist today.

Keeping the findings in our own notes instead.

Issue #182, opened 2026-06-15
MarkPad 2.6.11 AppImage does not start on Linux.


heyobscion:
## 🐞 Bug Report

**Describe the bug**  
MarkPad 2.6.11 AppImage does not start on Linux. The application immediately exits with an EGL error and no GUI window appears. The issue happens consistently on my system.

**To Reproduce**  
1. Download `Markpad_2.6.11_amd64.AppImage`
2. Make it executable:
   ```bash
   chmod +x Markpad_2.6.11_amd64.AppImage
   ```
3. Run the AppImage:
   ```bash
   ./Markpad_2.6.11_amd64.AppImage
   ```
4. See error in terminal and app exits immediately

**Expected behavior**  
The application should launch normally and open the MarkPad editor window.

**Screenshots**  
Not applicable (application fails to launch before any UI appears).

**Device**
- OS: CachyOS (Arch Linux based)
- Desktop: GNOME
- Session: Wayland (also tested on X11)
- GPU: Intel Arc Graphics (Mesa driver)
- App Version: v2.6.11 (Markpad_2.6.11_amd64.AppImage)

**Additional context**  
The application fails with:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

When extracting the AppImage and running the binary directly, an additional error appears:

```
Unable to spawn child process:
././/lib/x86_64-linux-gnu/webkit2gtk-4.1/WebKitNetworkProcess
(No such file or directory)
```

This suggests that WebKitGTK subprocess paths inside the AppImage are broken or incorrectly packaged.

I also tested multiple workarounds without success:
- `--disable-gpu`
- `--no-sandbox`
- `GDK_BACKEND=x11`
- `LIBGL_ALWAYS_SOFTWARE=1`
- `WEBKIT_DISABLE_COMPOSITING_MODE=1`
- `WEBKIT_DISABLE_DMABUF_RENDERER=1`

System graphics stack is working correctly (Mesa Intel Arc), and other GTK/WebKit applications run normally.

alecdotdev:
any success with 2.7.0?

PathGao:
@heyobscion — 2.7.0 will not have helped, and now we know why rather than guessing.

[...]

PR #463, opened 2026-08-05
fix(ci): unpin WebKitGTK so the AppImage bundles a Mesa-compatible version

LargeModGames:
## Symptom
The Linux AppImage exits immediately on Intel/Mesa systems with:

    Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...

No window ever appears. Reported in #182 (CachyOS, Intel Arc, Mesa); reproduced
on the current v2.7.0 release (Arch, Intel, Mesa 26.1.6, Wayland).

## Root cause
The release job in `.github/workflows/build.yml` pins the WebKitGTK stack:

    libwebkit2gtk-4.1-0=2.44.0-2
    libwebkit2gtk-4.1-dev=2.44.0-2
    ...

so `tauri build` bundles WebKitGTK 2.44.0 into the AppImage, and 2.44.0 fails EGL
initialization against modern Mesa. The abort comes from the bundled
`libwebkit2gtk-4.1.so.0` itself: forcing the app onto a current system WebKit
(2.52.x) makes it disappear. The usual env workarounds
(`WEBKIT_DISABLE_DMABUF_RENDERER=1`, `WEBKIT_DISABLE_COMPOSITING_MODE=1`,
`LIBGL_ALWAYS_SOFTWARE=1`) do not help.

## Fix
Drop the `=2.44.0-2` version pins so the ubuntu-24.04 runner installs the current
patched WebKitGTK from noble-updates (2.52.3-0ubuntu0.24.04.1 at time of writing),
which `tauri build` then bundles. All six packages are binaries of the one
`webkit2gtk` source published in lockstep, so they resolve to a single version.

If you'd rather keep an explicit pin, `=2.52.3-0ubuntu0.24.04.1` works equally;
I unpinned so it won't go stale again.

## Verification
Rebuilt v2.7.0 with `npm run tauri build -- --bundles appimage`:
- Before (bundled 2.44.0): `EGL_BAD_PARAMETER`, process exits, no window.
- After (bundled 2.52.x): launches and renders the full UI.

Environment: Arch Linux, Intel, Mesa 26.1.6, Wayland. Confirmed with the real
repackaged AppImage, not just in theory.

## Note
The deeper fragility is that the AppImage bundles WebKitGTK at all; an old bundled
WebKit against a newer host Mesa is inherently brittle. This CI change is the clean
app-side fix; the underlying behavior lives upstream in tauri-bundler.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -88,12 +88,12 @@ jobs:
           librsvg2-dev \
           patchelf \
           rpm \
-          libwebkit2gtk-4.1-0=2.44.0-2 \
-          libwebkit2gtk-4.1-dev=2.44.0-2 \
-          libjavascriptcoregtk-4.1-0=2.44.0-2 \
-          libjavascriptcoregtk-4.1-dev=2.44.0-2 \
-          gir1.2-javascriptcoregtk-4.1=2.44.0-2 \
-          gir1.2-webkit2-4.1=2.44.0-2
+          libwebkit2gtk-4.1-0 \
+          libwebkit2gtk-4.1-dev \
+          libjavascriptcoregtk-4.1-0 \
+          libjavascriptcoregtk-4.1-dev \
+          gir1.2-javascriptcoregtk-4.1 \
+          gir1.2-webkit2-4.1
           sudo apt-get install -y libgtk-3-dev libwebkit2gtk-4.1-dev libappindicator3-dev librsvg2-dev patchelf rpm xdg-utils \
           libxcb1-dev libxcb-render0-dev libxcb-shape0-dev libxcb-xfixes0-dev
 


PathGao:
This looks right to me, and the verification is the kind that settles it — a real repackaged AppImage on real hardware, with the usual `WEBKIT_DISABLE_*` workarounds ruled out rather than assumed. Four things I can add from the repo side, none of which contradict you.

**The pin has no recorded rationale.** It came in with `d19ae44` ("Add deps for AppImage", 2026-02-28) and the commit message gives no reason for `=2.44.0-2` — it reads like a copy of the Tauri CI recipe that was circulating at the time, not a deliberate constraint. So there is nothing to preserve by keeping it.

**The step already installed WebKitGTK twice, once pinned and once not.** The `sudo apt-get install -y libgtk-3-dev libwebkit2gtk-4.1-dev …` line immediately below the pinned block has always been unpinned. Your change makes the two consistent; before it, which version you got depended on apt resolution order between two lines in the same step. Worth mentioning in case it explains why this went unnoticed for so long.

**The blast radius is AppImage-only.** The `.deb` declares `Depends: libwebkit2gtk-4.1-0` — it links the host's WebKitGTK rather than carrying one — so `.deb` and `.rpm` users were never exposed to this. Only the AppImage bundles the engine, which is what makes it the one artifact coupled to the host's Mesa.

**On pin vs. unpin**, I'd take the unpin you have rather than moving to `=2.52.3-0ubuntu0.24.04.1`. The failure came *from* the pin going stale; a newer pin just sets the same alarm for two years out, and nobody will be watching that clock either. An unpinned build that breaks visibly in CI is easier to notice than a pinned one that quietly ships an obsolete engine.

Your closing note is the part I'd want the maintainer to read twice: this is the app-side fix, and the structural problem — an AppImage carrying a browser engine that has to agree with the host's graphics stack — is upstream in `tauri-bundler`. Unpinning moves that goalpost rather than removing it, so the same class will come back when Mesa moves past 2.52. Worth its own issue so it isn't rediscovered from a user report next time.

This should also close #182, whose reporter has been asked to retest on 2.7.0 and would presumably rather retest on something that starts.

PR #474, opened 2026-08-06
fix(scroll-sync): map the panes by source line instead of by ratio

PathGao:
Closes the surviving half of #205 — the end-of-document drift.

#316 removed Monaco's scroll-past-last-line padding from the sync range, which was the largest part of it, but the mapping was still proportional: a position became a share of the sending pane's scroll range and was read off the receiving pane's.

### Why a ratio cannot be right

Forty source lines of table render as one tall block; forty of prose render as a short one. Equal progress through the source is not equal progress through the rendered height, so the two panes' shares mean different things, and the difference accumulates. That is why what gets noticed is the end of the document.

### What replaces it

A `ScrollSyncPosition` now also carries the **source line** at the top of the sending pane's viewport, and each pane resolves that line through its own layout.

* **preview → editor** — the preview descends its `data-sourcepos` blocks to the narrowest one covering `scrollTop`, interpolates the line inside it, and Monaco reveals that line. Inverting Monaco's `getTopForLineNumber` by binary search keeps it correct with folded regions, wrapped lines and view zones, none of which have a uniform height.
* **editor → preview** — the top visible line goes to `findNearestAnchorElement`, and `getAnchorScrollTop` (the interpolation the tab restore already used) turns it into an offset.

Both directions are the *same* descent over the same tree with the same skip rules, so line → pixel → line is an identity rather than an approximation.

### Fallback

The front-matter carve-out is untouched and is now also the fallback. Front matter renders as a panel with no source range at all, so a front-matter position carries no line and travels as `section`+`ratio` exactly as before; a body position falls back the same way when the preview holds no annotated block.

A line no block *owns* — a blank line between blocks, or a position past the last one — resolves to the **nearest** block rather than falling back. Blank lines are about a third of a typical document, and falling back on each of them would put the panes back on the proportional mapping every time the reader crossed one.

### No feedback loop

The existing `isProgrammaticScroll` / `isApplyingExternalScroll` flags stop the echo. What is asserted here is that the mapping could not sustain a loop if an echo did get through: A → B → A is **idempotent** — it snaps once, by at most the margin between two blocks, onto a position that is its own image, and both components already ignore a move under 5px.

Getting there needed one fix a browser found and the DOM shim could not. Blocks touch — margins collapse, so one block's bottom edge is the next one's top — and letting both claim that shared pixel made the earlier one win, so a line mapped back to a pixel landed on a block top that the block *above* then claimed. Every echo walked the reader one block up the document. A block now owns `[top, bottom)`.

Both `setScrollTop` targets are also clamped into range before the 5px threshold check: a line near the end resolves to an offset a pane cannot reach, and an unreachable target fires no scroll event, which would leave the flag to be spent swallowing the reader's *next* real scroll.

### Cost — measured, not asserted

Chrome, over this pipeline's own output:

| document | offset → line | line → offset |
|---|---|---|
| 13,356 lines / 320,000px / 6,709 blocks | 0.89 ms | 0.28 ms |
| 1,757 lines / 43,000px | 0.21 ms | 0.12 ms |
| 437 lines / 11,000px | 0.17 ms | 0.10 ms |

Source ranges and derived container spans are memoised in `WeakMap`s keyed on the element; without them the same walk is ~6 ms. **Nothing invalidates them explicitly** — the preview is rebuilt wholesale by `bind:innerHTML`, so a changed document is a new set of nodes and the entries describing the old ones are collected with them. Folding, task checkboxes and rich-content rendering all leave the annotated blocks and their ranges where they were.

**The scroll handler got cheaper overall, not dearer.** Capturing the tab's reading position was a separate `querySelectorAll('[data-sourcepos]')` over the whole preview plus `offsetTop`/`offsetHeight` on every element it returned, on the same event — 8.3 ms on the 13,356-line document. It now shares the one descent, which also settles two disagreements it had with the restore: it took the *outermost* element covering an offset where `findAnchorElement` takes the narrowest, and it had no opinion about the boxless `<br>` of #464.

### Trade-off worth naming

What the mapping aligns is the line at the **top** of each viewport, which is the only thing two panes of different heights can agree on. When the editor is scrolled to its last line the preview is therefore not necessarily at its own last pixel — the same lines take more height there. The top lines agree exactly, which is the alignment the report is about.

### Tests

`scripts/scrollSyncBlockMapping.test.ts` drives the mapping in both directions over real `processMarkdownHtml` output and measures the disagreement in **source lines**, with the pre-fix mapping kept alongside as the control:

```
ratio mapping: worst drift 54.9 source lines   (tail of the document)
block mapping: worst drift 0.9 source lines
ratio mapping: worst drift 64.8 source lines   (whole document)
block mapping: worst drift 0.9 source lines
preview -> editor: worst drift 0.0 source lines
editor -> preview -> editor: first echo up to 17.8px, second 0.000px
```

The shim has no layout at all — `grep -rn offsetTop scripts/` returns nothing, which is why #464's defect survived a suite scoring 100% — so the layout is injected through the same `measure` callback the browser fills with `offsetTop`. What that **cannot** model is whether a browser really lays a table out the way the fixture's heights say; the property under test only needs rendered height not to be proportional to source lines, which is the premise of the bug. The touching-block case gets a second fixture, because the first one's synthetic margins hid it and only Chrome found it.

Falsification — mapping reverted, tests kept:

```
✖ the tail of the document stays on the same source line
  AssertionError: the preview drifted 54.9 source lines from the editor at scrollTop 7295 (of 12158)
✖ every part of the document stays on the same source line, not only the tail
  AssertionError: the preview drifted 64.8 source lines at scrollTop 5775
✖ an echo through the other pane settles in one step
  AssertionError: a second echo would move the editor a further 1238.5px at scrollTop 4985
✖ repeating the echo does not walk the panes down the document
  AssertionError: twenty echoes moved the editor 1841.0px
ℹ pass 9   ℹ fail 10
```

And with only the boundary-pixel fix reverted:

```
✖ touching blocks do not walk the panes up the document
  AssertionError: a second echo moved the preview 38.0px between touching blocks
✖ the pixel two blocks share belongs to the lower one
```

### Verification

`npm audit` (0 vulnerabilities), `npm run check` (0 errors), `npm test` (702 passing), `cargo test` (157 passing).

---

@Guardiancelte — this is the per-block mapping I mentioned. Your two stress-test files are exactly the shape it is meant to fix (the fixture here is modelled on them: prose for most of the document, then tables and figures near the end). Would you retest both once this lands?

### DIFF
--- a/scripts/scrollSyncBlockMapping.test.ts
+++ b/scripts/scrollSyncBlockMapping.test.ts
@@ -1,0 +1,1004 @@
+/**
+ * Split-view scroll sync: does the pane the reader is not touching show the
+ * same part of the document?
+ *
+ * The mapping used to be a ratio — the sending pane's share of its own scroll
+ * range became the receiving pane's share of its. That cannot be right in
+ * general. Forty source lines of table render as one tall block and forty of
+ * prose as a short one, so equal progress through the source is not equal
+ * progress through the rendered height, and the error accumulates towards the
+ * end of the document. That is the surviving half of #205: the reporter sees
+ * the panes "quite a few lines off" at the end of a document full of tables.
+ *
+ * So the unit of the measurement here is a SOURCE LINE, not a pixel. Every test
+ * drives the real mapping in one direction and then asks the receiving pane
+ * which source line it is now showing:
+ *
+ *   editor scrollTop -> line -> ScrollSyncPosition -> preview scrollTop -> line
+ *
+ * `RATIO_ONLY` re-runs the same sweep with the `line` field stripped out of the
+ * position, which is exactly the pre-fix mapping, and is kept as the control:
+ * if the block mapping ever collapses back into a proportional one the two
+ * drift figures collapse into each other.
+ *
+ * ---------------------------------------------------------------------------
+ * WHAT THE DOM SHIM CANNOT MODEL, AND WHAT IS DONE ABOUT IT
+ *
+ * `renderProtocolDom.ts` has no layout whatsoever — `grep -rn offsetTop
+ * scripts/` finds nothing — which is why #464's defect (anchors resolving to a
+ * boxless `<br>`) survived a suite that scored 100%. A mapping from pixels to
+ * lines is nothing BUT layout, so a test that only walked the shim would be
+ * measuring the same nothing.
+ *
+ * The layout is therefore injected. `layOut` below assigns every rendered block
+ * a top and a height, bottom-up, and `getSourceLineAtPreviewOffset` /
+ * `getPreviewOffsetForSourceLine` take that as the `measure` callback — the
+ * same parameter the browser fills with `offsetTop` / `offsetHeight`. What is
+ * asserted is the mapping's arithmetic and its choice of element over real
+ * `processMarkdownHtml` output; what is NOT asserted is that a browser lays a
+ * table out the way `BLOCK_HEIGHTS` says. The property under test does not
+ * depend on the specific numbers, only on rendered height not being
+ * proportional to source lines, which is the premise of the bug.
+ */
+
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { installShimDom, parseHtml, NODE_ELEMENT, type ShimElement } from './renderProtocolDom.ts';
+
+installShimDom();
+
+const { processMarkdownHtml } = await import('../src/lib/utils/markdown.ts');
+const { getPreviewOffsetForSourceLine, getSourceLineAtPreviewOffset } = await import(
+	'../src/lib/utils/previewAnchor.ts'
+);
+const {
+	getLineAtVerticalOffset,
+	getScrollSyncPositionFromPixels,
+	getScrollTopForSyncPosition,
+	getVerticalOffsetForLine,
+} = await import('../src/lib/utils/scrollSync.ts');
+
+type AnchorNode = Parameters<typeof getSourceLineAtPreviewOffset>[0];
+type ScrollSyncPosition = Parameters<typeof getScrollTopForSyncPosition>[0];
+
+const FILE_PATH = '/documents/stress.md';
+
+/* ------------------------------------------------------------------ */
+/* a document whose rendered height is not proportional to its source  */
+/* ------------------------------------------------------------------ */
+
+class DocumentBuilder {
+	private line = 1;
+	private readonly parts: string[] = [];
+
+	private push(html: string, lines: number) {
+		this.parts.push(html);
+		this.line += lines;
+	}
+
+	blank() {
+		this.line += 1;
+	}
+
+	heading(level: number, text: string, slug: string) {
+		const line = this.line;
+		this.push(
+			`<h${level} data-sourcepos="${line}:1-${line}:${text.length + level + 1}">` +
+				`<a href="#${slug}" aria-hidden="true" class="anchor" id="${slug}"></a>${text}</h${level}>`,
+			1,
+		);
+	}
+
+	paragraph(text: string) {
+		const line = this.line;
+		this.push(`<p data-sourcepos="${line}:1-${line}:${text.length}">${text}</p>`, 1);
+	}
+
+	/** Three source lines — header, separator, one row — for a whole rendered table. */
+	table(rows: number, label: string) {
+		const start = this.line;
+		const end = start + rows + 1;
+		const body = Array.from(
+			{ length: rows },
+			(_, index) => `<tr><td>${label} ${index}</td><td>value ${index}</td></tr>`,
+		).join('');
+		this.push(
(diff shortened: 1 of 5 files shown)


Guardiancelte:
Your explanation goes way above my head but yes will test once available :).

Guardiancelte:
@PathGao this is a lot better! I dont know if you prefer me replying here or in the original issue so I will do both. 

(#304) fully fixed the jumping around. 

(#196) and (#316) have improve the scrolling by a large margin. 

Basically the easy stress test pass without issues. 

For the hard one. It is a lot better and at that point it depend how much time you want to spend on this :). 

At this point it might also be the desired behaviour as well.

===========================

It looks like depending if having on the mouse on the left side or the right side impact the calculation differently 

Screenshot with mouse on the edit panel side, the only diffence is I scrolled one click down :
Image
because diagrams are being used I am guessing it is messing with the calculation would be my guess, but as you can see the top title on the right is # 11. Mermaid Diagrams but on the right is Sequence Diagram
Image
If the calculation are running based left side when I have my mouse of the left side, I would expect top titles to align. 

But in the exact same location with my mouse on the right side, it smoothly move the right side. 

Image

Image

============================

and you get more issues for large tables, smaller tables seem fine:
mouse one the left side, one scroll down, it sends me to the top of the document, and it never comes back until I am passed both tables

Image
top of the doc post one scroll down
Image

mouse to the right side, the table display on the left, I am using scrolling up as I think it shows the issues better
start with the large table just picking out on the right side, that move the bottom of the table on the left in view

Image

then the left side is frozen until the top of the table is visible on the right side (have to scroll a lot), basically jumping a 100 lines or so. and when that unfreeze, one scroll down on the right side shot the left side all the way back down

Image

========================

A similar issue with on scroll down jumping is happening here but not sure why:
mouse on the left side, picture are taken as one scroll interval (4 scroll total), sending me in 3 seperate places in the doc:

Image

Image

Image

Image

Image

=======================

I am running markpad is on that screen with zero in app zoom, that might impact scroll numbers or something, so making you aware

Image

PathGao:
@Guardiancelte — thank you for retesting this properly, and for the screenshots. Both of your cases were real, and neither of them was the mapping I described above: they were the *measurement* underneath it. #515 has the fix.

The short version, without the arithmetic. To place a block, the code asked the browser "how far down are you?" via `offsetTop`. That property does not answer "from the top of the document" — it answers "from my **offset parent**", and the browser makes a `<table>` the offset parent of its own rows, whatever anything else says. Measured in Chrome on this app's stylesheet: a table row sitting 995px down the document reports `1`. A code block reports `0`, because the copy button needs a positioned wrapper around it.

Everything you saw follows from those two numbers being read as positions in the document:

- **Big tables, one scroll down → top of the document.** A line inside the table maps to ~200px, so the pane goes to the top. Small tables did it too — you just could not see it, because ~200px into a small table is nearly where you already were.
- **The other pane frozen until you are past the table.** Same numbers, other direction: every row's box is *far above* where you actually are, so the nearest row is always the last one. The pane sits on the end of the table, then snaps.
- **Mouse on the left vs on the right behaving differently.** Those are the two directions of the same defect, and they fail differently. Not the mouse — which pane is sending.
- **Mermaid.** A different one, same family: the rendered diagram threw away the code block's line numbers, so the diagram was invisible to the mapping and its height got charged to the block next to it. That is your `# 11. Mermaid Diagrams` against `Sequence Diagram` — almost exactly one diagram's worth of disagreement.

Your in-app zoom is not involved.

What I would ask, once #515 is in a build: your **large-table** file, mouse on each side in turn, and the Mermaid one for the headings. If something still drifts I would rather hear it than not — the last round is the reason this one exists.

One thing that is *not* a bug, so you can tell it apart from one: what the panes align is the line at the **top** of each pane. At the very bottom of a document the preview will often still have room to scroll while the editor has none. The top lines agreeing is the alignment; the bottoms are not meant to.

Guardiancelte:
Side note. You seem to have a very good stack for LLM coding and also GitHub interactions. I am guessing considering number of commits and speed of feedback. 
Do you mind if I ask what it is?

PathGao:
> Side note. You seem to have a very good stack for LLM coding and also GitHub interactions. I am guessing considering number of commits and speed of feedback. Do you mind if I ask what it is?

"Thanks! My current LLM setup is pretty straightforward:

Claude Max (20x): Got it initially for Claude Science, but the harness was a headache. Claude Code, however, is fantastic—I'm using it actively to patch open-source projects I use daily.

Pi + DeepSeek v4 Pro (via pi-web): Replaced Codex with this recently (Codex just wasn't cutting it). It’s super fast, minimalist, dirt cheap, and surprisingly effective.

Still benchmarking and tweaking the workflow, but it’s been working great so far!"

On a side note, I actually just started contributing PRs on GitHub about two weeks ago, so I’m still learning the ropes on the git/GitHub side of things (getting more comfortable every day, though!). Using LLMs has been a game-changer—it lets me focus entirely on software design and core principles, which helped me pick things up super fast."

PR #494, opened 2026-08-06
docs(releasing): close the loophole in the redirect rule before someone finds it

PathGao:
Follow-up to #485, which added `RELEASING.md` §5 — the rule that nothing may occupy `alecdotdev/Markpad`, because a repository or fork there permanently voids the transfer redirect that every install up to v2.7.0 depends on for updates.

The rule is right and I am not weakening it. What is missing is the answer to the question a reader will actually ask.

**The loophole.** The redirect is only the *means*. What those installs need is for that URL to keep returning a current `latest.json`, whose contents are absolute download URLs verified against the `pubkey` compiled into the binary. So a repository at the old location could keep them working — publish a release there whose `latest.json` names this repository's artifacts, same signing key, and old clients verify it and update normally.

Someone will notice this, and they will notice it at the worst moment, which is while they are doing whatever made them want the old name back.

**Why it is a bad trade, stated so nobody has to work it out under pressure.** It converts a guarantee that costs nothing into an obligation that must be met on every release, forever, by whoever happens to be releasing. The first time it is missed, old clients are handed a stale feed offering a version they already have — they stop updating exactly as if the URL had 404'd, except now nothing looks broken to anyone.

So the section now says the exception exists and why not to take it, rather than leaving a reader to find it alone.

## Verification

`npm test` passes, unchanged. Nothing else is touched — this is one inserted paragraph in `RELEASING.md`.

**Not verified:** I have not tested what GitHub actually does if a repository is created at a transferred location — the claim that the redirect is voided is from [GitHub's documentation](https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository), quoted in the section, not from an experiment. The same is true of the claim that a `latest.json` served from the old location would satisfy `tauri-plugin-updater`; that follows from reading the updater's `get_urls` and our feed's shape, and has not been run.


PathGao:
Closing this in favour of a check.

The point @PathGao's reviewer made is that §5 keeps needing more words because it is not adjacent to anything. §4 earns its place by sitting directly after the step it constrains — generating the keypair is step 1, "the pubkey is permanent" is the warning on step 1. §5 constrains an action that is not a step in this runbook at all: nobody cutting a release is going to fork the repository. Adding another paragraph, as this PR does, treats the symptom.

More to the point, the loophole this PR spends a paragraph closing does not need prose. A check that asserts the old URL still returns a feed pointing at this repository tests the property we actually care about, rather than the proxy of whether a repository exists there — so a repository at the old location that *does* serve a correct feed passes, correctly, and one that does not goes red. The exception polices itself.

Replacing this with that check, and shortening §5 rather than growing it.

Issue #498, opened 2026-08-06
AppImage 2.7.1 shows a blank window: bundled libwayland-client.so.0 breaks EGL on newer Mesa


LargeModGames:
## Summary

The v2.7.1 Linux AppImage still fails to display on hosts whose Mesa is newer than the ubuntu-24.04 build runner. The window opens but stays blank white; the WebKit WebProcess aborts during EGL initialization with:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

This is a second, independent cause behind #182. The WebKitGTK version pin from #463 did land (verified below); a separate packaging problem remains.

## This is not the #463 problem (that fix landed)

A dlopen probe of the library inside the published `Markpad_2.7.1_amd64.AppImage`, calling `webkit_get_major/minor/micro_version`, reports:

```
BUNDLED_WEBKITGTK_VERSION = 2.52.3
```

So #463 worked: the AppImage now bundles a modern WebKitGTK, not the old 2.44.0. The remaining failure is not about the WebKit version.

## Root cause

The AppImage bundles `usr/lib/libwayland-client.so.0` (the ubuntu-24.04 copy). On a host with a newer libwayland/Mesa, that older bundled copy shadows the host one. The host `libEGL` (libglvnd, which links libwayland-client) then fails `eglGetPlatformDisplay` with `EGL_BAD_PARAMETER`, so WebKit's WebProcess cannot create an EGL display and aborts. The GTK shell window survives, which is why the result is a blank window rather than an immediate exit.

`libwayland-client.so.0` is on the standard AppImage excludelist (AppImageCommunity/pkg2appimage) precisely because it is coupled to the host graphics stack and must never be bundled.

Verbose Mesa logging (`EGL_LOG_LEVEL=debug`, `LIBGL_DEBUG=verbose`, `MESA_DEBUG=1`) prints nothing before the abort, which places the rejection at the EGL dispatch entry, before any DRI driver is reached. That is consistent with a libwayland ABI/version mismatch rather than a GPU or driver fault.

## Evidence: single-library isolation

Extracted the published AppImage, moved individual libraries out of the AppDir so the host copy is used, then ran the app on the default (forced x11) backend:

| Removed from AppDir | EGL init | WebProcess | Result |
| --- | --- | --- | --- |
| nothing (as shipped) | EGL_BAD_PARAMETER | aborts | blank window |
| only `libwayland-client.so.0` | ok | alive | **full UI renders** |
[...]

Issue #496, opened 2026-08-06
2.7.1 is ready to cut


PathGao:
`package.json` is already at 2.7.1 and 22 commits have landed since v2.7.0. This
is what is in it, what needed a decision and got one, and one thing about the
old repository URL that has to be said out loud because nothing in the code can
say it.

---

## #466 — the installer and uninstaller are gone

You asked whether we could strip the install/uninstall code and let NSIS own it.
#481 does that: `setup.rs` (1684 lines), `Installer.svelte` (548) and
`Uninstaller.svelte` (248) are deleted, and `mslnk` and `chrono` go with them —
`setup.rs` was their only user.

**Why this clears the VirusTotal detection.** The uninstall path wrote a batch
file to `%TEMP%`, wrote a `.vbs` that ran it with a hidden window and then
deleted itself, and launched the pair through `wscript.exe`. Hidden script →
self-delete → launch an executable is close to the definition of
`Trojan:Win32/Wacatac.B!ml`, which scores behaviour rather than matching a
signature. Every property had a reason — the batch waited on the PID rather than
the image name so a portable copy survives an uninstall (#358) — but the shape
is what gets scored, so the only thing that clears it is removing the shape.

**The install half was already dead.** `get_app_mode` entered installer mode only
on a filename containing `installer` or an explicit `--install`, and #355 stopped
publishing any installer-named artifact. Nothing shipped could reach it.

**The uninstall half needed a migration, and it turned out not to need a Windows
box.** I had asked you to check on Windows whether the updater's `/UPDATE` run
still rewrites the registry entry. That question does not arise. Reading the
bundler template at the version that built 2.7.0 (`tauri-cli-v2.9.6`):

```
Section Install = lines 614..717
  line 682   WriteRegStr SHCTX "${UNINSTKEY}" "UninstallString" "$INSTDIR\uninstall.exe"
  occurrences of $UpdateMode inside 614..717:  0
```

The write is not behind a mode check at all — the `$UpdateMode` guards in that
file sit in `Section WebView2` and `Section Uninstall`. So it lands whether or
not `/UPDATE` is passed. Both sides also address the same key: the custom
installer used `APP_NAME` and the template uses `${PRODUCTNAME}`, both `Markpad`.

Two backstops cover the gap that remains. `Section Install` writes through
`SHCTX`, so an old install that chose the other registry hive keeps a stale
Add/Remove Programs row; a hook in `hooks.nsi` deletes it, matching on the
[...]

PR #499, opened 2026-08-06
fix(ci): strip host-coupled libwayland from the AppImage so it starts on newer Mesa

LargeModGames:
## Problem

The Linux AppImage still fails to display on hosts whose Mesa is newer than the ubuntu-24.04 build runner (#498). The window opens but stays blank white and WebKit's WebProcess aborts:

```
Could not create default EGL display: EGL_BAD_PARAMETER. Aborting...
```

This is a second, independent cause behind #182, distinct from the WebKitGTK version pin fixed in #463. The published v2.7.1 AppImage bundles WebKitGTK 2.52.3 (confirmed by dlopening the bundled library and reading `webkit_get_*_version`), so #463 landed correctly.

## Root cause

The AppImage bundles `libwayland-client.so.0` from the ubuntu-24.04 runner. On a host with a newer libwayland/Mesa, that older bundled copy shadows the host one, and host `libEGL` then fails `eglGetPlatformDisplay` with `EGL_BAD_PARAMETER`.

Single-library isolation (extract the published AppImage, move one library out of the AppDir at a time, run): removing only `libwayland-client.so.0` is necessary and sufficient to make the full UI render. It is on the AppImageCommunity/pkg2appimage excludelist precisely because it must come from the host.

The failure is backend independent (the default forced `GDK_BACKEND=x11` and `GDK_BACKEND=wayland` both abort), and `WEBKIT_DISABLE_DMABUF_RENDERER`, `WEBKIT_DISABLE_COMPOSITING_MODE`, and `LIBGL_ALWAYS_SOFTWARE` do not help.

## Change

A step after "Build Linux" that extracts the AppImage, removes the six host-coupled graphics libraries (`libwayland-*`, `libxcb-render/shm`), repacks with appimagetool, and re-signs.

The re-sign is required: `tauri build` already signed the pre-strip file, and the updater verifies the downloaded AppImage against the signature in `latest.json`, so the repacked artifact must be signed again. Both the upload step and the `latest.json` job read the regenerated `.AppImage.sig`.

`--appimage-extract` and `appimagetool --appimage-extract-and-run` are used so no libfuse2 is needed on the runner.

## Validation

Validated locally on Arch (Intel, Mesa 26.1.6): extracting the published v2.7.1 AppImage, removing the libs, repacking, and launching the result renders the full UI with no EGL error (before/after screenshots on #498). The signer invocation and the no-FUSE extract/repack were exercised directly.

Not validated, and it cannot be before a release: the step running on the ubuntu-24.04 runner with the real signing secret. As with #463, the resulting AppImage is the first artifact that can be tested, so it is worth confirming the published file starts before announcing.

## Why the #463 verification did not catch this

My #463 check rebuilt on Arch, so linuxdeploy bundled Arch's `libwayland-client` (matching the host) and could not reproduce the mismatch. The ubuntu-24.04 release build bundles the older one. Two independent bugs: the version pin (#463) and this libwayland bundling.

Addresses #498 and the remaining AppImage failure in #182. Leaving both open so they can be closed after the published 2.7.x AppImage is confirmed to start.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -232,6 +232,72 @@ jobs:
           TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY_PASSWORD }}
         run: npm run tauri build
 
+      # tauri/linuxdeploy bundles libwayland-client.so.0 (and other host-graphics
+      # libraries) from the ubuntu-24.04 runner into the AppImage. On a host whose
+      # Mesa is newer than the runner's, that older bundled libwayland-client
+      # shadows the system one and breaks host libEGL: WebKit's WebProcess aborts
+      # with "Could not create default EGL display: EGL_BAD_PARAMETER" and the
+      # window comes up blank. These libraries are on the AppImage excludelist
+      # precisely because they must come from the host. Strip them, repack, and
+      # re-sign so the updater signature still matches. See #498; #463 fixed the
+      # separate WebKitGTK version pin.
+      - name: Strip host-coupled libraries from AppImage
+        if: matrix.platform == 'ubuntu-24.04'
+        env:
+          TAURI_SIGNING_PRIVATE_KEY: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY }}
+          TAURI_SIGNING_PRIVATE_KEY_PASSWORD: ${{ secrets.TAURI_SIGNING_PRIVATE_KEY_PASSWORD }}
+        run: |
+          set -euo pipefail
+
+          APPIMAGE=$(find src-tauri/target/release/bundle/appimage -maxdepth 1 -name '*.AppImage' | head -1)
+          if [ -z "$APPIMAGE" ]; then
+            echo "::error::No AppImage found to repackage" >&2
+            exit 1
+          fi
+          APPIMAGE=$(realpath "$APPIMAGE")
+          echo "Repackaging $APPIMAGE"
+
+          # appimagetool + a type2 runtime. --appimage-extract-and-run (below) and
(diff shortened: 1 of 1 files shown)


PathGao:
Thank you for testing the built artifact rather than trusting the diff — including mine. I wrote in #496 that #182 would be fixed the moment 2.7.1 shipped, on the strength of reading #463. You went and ran it. That was the right order and I have corrected the issue.

**The Linux judgement here is yours, and I am not going to second-guess it.** I have no Linux hardware and cannot reproduce or verify any of it, and it would not be a good trade to hold up a fix from someone who can run it on a reviewer who cannot. The single-library isolation table settles the cause about as directly as it can be settled.

One thing I do want to flag, because it is not a Linux question and does not need a Linux machine to judge:

```
wget .../appimagetool/releases/download/continuous/appimagetool-x86_64.AppImage
wget .../type2-runtime/releases/download/continuous/runtime-x86_64
```

`continuous` is a mutable tag, and this step runs in the job that has `TAURI_SIGNING_PRIVATE_KEY` and `TAURI_SIGNING_PRIVATE_KEY_PASSWORD` in its environment — the step even re-signs with them, correctly. So whatever is behind those two URLs at build time executes next to the project's signing key. If either release were ever replaced, that is the blast radius.

Pinning to a tagged release, or keeping `continuous` and checking a recorded `sha256sum` before `chmod +x`, both close it. Your call which; it is a few lines either way and it does not change what the fix does.

For context on why this reads as load-bearing here: `RELEASING.md` §4 spells out that the pubkey is permanent — if that key is ever compromised, a new keypair has to be generated and every user has to reinstall by hand. There is no revocation path.

Everything else — the six libraries, the repack mechanics, the backend independence — I defer to you on entirely.

PR #516, opened 2026-08-07
docs(samples): add the scroll-sync stress document, by @Guardiancelte

PathGao:
The document [@Guardiancelte](https://github.com/Guardiancelte) has been retesting scroll sync against, kept in the repository so the next change to it has the same thing to fail against.

3,164 lines — 124 headings, 543 table rows, three Mermaid diagrams, code blocks in eight languages, footnotes, math, raw HTML and text in seven scripts. It is what found both defects fixed in #515: a table's rows and a rendered diagram report positions the scroll mapping could not read, and neither showed up in a fixture written by the person fixing the bug.

**The commit is authored to him**, through his GitHub noreply address, because he wrote it — this is his contribution, not ours carrying his file. @Guardiancelte, if you would rather it were not committed here, or not under your name, say so and it comes straight back out.

He also reported #205 and retested every round of fixes against it: #196, #304, #316, #474, and now #515.

### DIFF
--- a/samples/stress-test-hard.md
+++ b/samples/stress-test-hard.md
@@ -1,0 +1,3165 @@
+---
+title: "Markdown Reader Stress Test"
+author: "OpenAI"
+date: "2026-07-21"
+version: "1.0"
+tags:
+  - markdown
+  - reader-test
+  - formatting
+---
+
+# Markdown Reader Stress Test
+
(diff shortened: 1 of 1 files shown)


Guardiancelte:
No worries, that is actually my first public commit then ;). Honoured that is for such a cool project.

PR #339, opened 2026-08-02
Add a minimal macOS Settings menu

PathGao:
## Summary

- reduce the macOS native menu to app-level actions
- add `Settings…` with the standard `Cmd+,` shortcut
- route Settings to the currently focused Markpad window

Fixes #281

## Validation

- `npm ci`
- `npm run check`
- `npm test`
- `cargo test`

## Local bundle validation

- macOS Bundle: `com.alecdotdev.markpad.pr281`, version `2.6.14-pr281.local.1`
- Built with `npm run tauri build` in an isolated clone.
- Verified the native menu contains only About, Settings…, Check for Updates…, Services, Hide, and Quit. `Settings…` opened the existing Settings modal in the focused window.

Issue #281, opened 2026-07-30
Discuss a minimal macOS application menu and shared Settings entry


PathGao:

PathGao:
@alecdotdev We hit this while validating the macOS build: Settings is readily available from the custom titlebar overflow, but users naturally look for it in the app menu first. We do **not** propose a full macOS File/Edit/View/Window hierarchy.

Our current design direction is deliberately narrower:

- the shared custom titlebar/Home menu remains the only surface for document, preview, tab, export, window-tag, and multi-window actions;
- the macOS app menu stays a small system-integration surface;
- it should include one `Settings…` item with `⌘,`, routed to exactly the same frontend action as the titlebar gear;
- About, Check for Updates, Services, Hide, and Quit remain there; no duplicated document commands.

This also matters for #234 (Settings/workbench) and the pending multi-window work (#279 / #280): keeping those actions out of the macOS menu avoids a second platform-specific command model.

Would you prefer `Settings…` (the current macOS wording) or `Preferences…`, and do you see any other app-global action that belongs in this intentionally minimal native menu?

PathGao:
@alecdotdev One missing part of the background: this came up while validating the pending window-organization UI. We are adding window tags and “Merge All Windows Here” to the shared Home menu below the export actions. That made the ownership of the two menus visible:

- the custom Home/titlebar menu is a cross-platform workspace menu, and it is where Markpad-specific document, export, preview, tag, and multi-window actions belong;
- the macOS app menu is a system menu, but it currently feels like a second, incomplete way to look for application commands. A macOS user will reasonably check it for Settings first, then has to discover a separate custom menu.

The concern is therefore not that the native menu is too small. It is that the boundary is implicit, so the two surfaces look like competing partial menus.

We considered three directions:

1. Mirror a full File/Edit/View/Window menu on macOS. This improves familiarity, but duplicates the shared UI, needs platform-specific enabled/checked state, and would be hard to keep correct as window/tag features evolve.
2. Keep no meaningful native app-menu entry points. This maximizes UI parity, but leaves a standard macOS destination for app-wide settings undiscoverable.
3. Keep a deliberately minimal native bridge (our recommendation): only app-global/system actions, with `Settings…` forwarding to the exact same handler as the shared titlebar gear. The workspace menu remains canonical for all document and multi-window operations.

[...]

PR #527, opened 2026-08-07
fix(macos): give the menu back the commands the web side cannot bind

### LINKED ISSUE #526: Pasting text to theme import field fails
### How did you install Markpad?

macOS — .dmg

### App version

2.7.1

### OS and version

MacOS Tahoe 26.5.2

### Linux only: distribution, session and GPU/driver

_No response_

### What happens

The input field beneath 'Input VS Code theme' no longer accepts pasted text
The full theme url must be entered manually, and cannot be pasted into the field after copying from the web browser.

### What you expected

Pasting text into the field should work using keyboard shortcuts

### Steps to reproduce

1. Open Settings
2. Navigate to Appearance tab
3. Click the Browse Themes button
4. Locate a theme and copy the url from vscodethemes.com
5. Click into the field
6. Use keyboard shortcut to paste (CMD + V)
7. Observe text is not pasted into the field
8. Right/Option click in the field
9. Observe missing paste option; no options available for pasting urls

### The document, if one triggers it

_No response_

### Screenshots

Image

### Anything else

_No response_

PathGao:
Fixes #526.

Stacked on #528 (dompurify pin) so CI can run — merge that one first.

## What happens

On macOS nothing can be pasted into a plain input — the theme-import field in Settings, the rename prompt — by keyboard or by right-click. ⌘M and ⌃⌘F do nothing either. All of it broke in 2.6.10 with #339.

## Why

#339 reduced the native menu to app-level actions and dropped File/Edit/Window, on the grounds that Markpad keeps its document actions in the in-window controls. That holds for document actions — `MarkdownViewer.handleKeyDown` really does bind ⌘N/⌘O/⌘S/⌘W/⌘±/⌘0/⌘Tab web-side, and those still work. It does not hold for commands the web side has no way to bind:

1. **Text editing.** macOS routes ⌘X/⌘C/⌘V/⌘A and undo through the main menu to the first responder — WKWebView declines them precisely so the menu can. With no Edit menu they go nowhere. Monaco was unaffected only because `Editor.svelte` implements its own ⌘V against the Rust clipboard commands, which is why this looked field-specific rather than app-wide.
2. **Window-server actions.** Minimize and full screen have no web API to bind at all. No menu item, no key equivalent, no shortcut.

Plus one that is not AppKit's doing: `handleContextMenu` `preventDefault()`s across the whole document and opens Markpad's preview menu, which has no edit items — so right-clicking a text field offers no Paste either. The modal backdrop does the same to the rename prompt.

## Changes

- **Edit menu**: Undo, Redo, Cut, Copy, Paste, Select All.
- **Window menu**: Minimize, Toggle Full Screen. Full Screen belongs in a View menu by convention; View would hold that one item and nothing else, so it rides in Window — a deliberate deviation, not an oversight.
- Text fields keep the webview's own editing menu, in the document handler and behind modals.

File stays out, and so do Hide Others / Show All — #281's intent is otherwise untouched.

Monaco keeps its ⌘V: it calls `preventDefault()`, WebKit reports the key as handled, and the menu never fires. Image paste and URL linkify still go through `Editor.svelte`.

## Validation

- `npm run check` 0 errors · `npm test` 843 pass · `cargo test` 141 pass
- Paste into the theme-import field by ⌘V and by right-click → Paste: confirmed by hand on macOS 26.
- ⌘M and ⌃⌘F measured on a dev build, before and after, by window size — `AXFullScreen` is unreliable for this window's overlay title bar:

  | | ⌃⌘F | ⌘M |
  |---|---|---|
  | without Window menu | 708x923 → 708x923 | not minimized |
  | with it | 708x923 → 1470x33 | minimized |

  Control for "did the keystroke arrive at all": ⇧⌘M, which is bound web-side, took the window count 1 → 2 in the same session.

- ⌘\` still cycles windows either way — it is a system binding, not a menu key equivalent — so no menu item is owed for it.
- Tests updated to state the new intent: `issue281MinimalMacosMenu.test.ts` still keeps File and the document shortcuts out, and now requires Edit and Window; `menuModalGuards.test.ts` asserts the backdrop consumes context menus *outside* text fields.

### DIFF
--- a/scripts/issue281MinimalMacosMenu.test.ts
+++ b/scripts/issue281MinimalMacosMenu.test.ts
@@ -18,8 +18,25 @@ test('macOS native menu keeps only application-level actions', () => {
 	assert.match(menuSetup, /PredefinedMenuItem::services\(app, None\)/);
 	assert.match(menuSetup, /PredefinedMenuItem::hide\(app, None\)/);
 	assert.doesNotMatch(menuSetup, /PredefinedMenuItem::(?:hide_others|show_all)/);
-	assert.match(menuSetup, /\.items\(&\[&app_submenu\]\)/);
-	assert.doesNotMatch(menuSetup, /SubmenuBuilder::new\(app, "(?:File|Edit|Window)"\)/);
+	assert.match(menuSetup, /\.items\(&\[&app_submenu, &edit_submenu\]\)/);
+	assert.doesNotMatch(menuSetup, /SubmenuBuilder::new\(app, "(?:File|Window)"\)/);
+});
+
+// Document actions stay in the in-window controls (#281), but Edit is not a
+// document menu: macOS routes ⌘X/⌘C/⌘V/⌘A through the main menu to the
+// first responder, so dropping it left every native input unable to paste
+// (#526).
+test('macOS menu keeps the standard Edit items so text fields can paste', () => {
+	const menuSetup = sliceBetween(
+		tauriLib,
+		'#[cfg(target_os = "macos")]\n            {\n                use tauri::menu',
+		'\n            let config_dir',
+	);
+
+	assert.match(menuSetup, /SubmenuBuilder::new\(app, "Edit"\)/);
+	for (const item of ['undo', 'redo', 'cut', 'copy', 'paste', 'select_all']) {
+		assert.match(menuSetup, new RegExp(`PredefinedMenuItem::${item}\\(app, None\\)`));
+	}
 });
 
 test('native Settings opens only the focused window settings modal', () => {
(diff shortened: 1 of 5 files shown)


PathGao:
Rebased onto #528. `npm audit` started failing on GHSA-55q2-fjhq-7xh7 (dompurify `<= 3.4.12`, published after master's last green run), which killed the `test` job before it reached `npm test` — nothing to do with this change, and it hits every open PR. #528 moves the pin; this PR now sits on top of it so its own checks can actually run.

Merge order: #528, then this one.

PR #622, opened 2026-08-12
perf(preview): debounce the rebuild against a pause, not against a keystroke

PathGao:
A stopgap, and labelled as one in the code.

### Why the number moved

The preview rebuild is **not pixel-identical** to the render it replaces, and that — not its cost — is what the reader sees.

`{@html}` destroys and recreates the whole article. Everything that settles afterwards settles a frame or two later: inline maths through KaTeX's auto-renderer, font metrics, the `ResizeObserver` remeasuring every fold. Any of them can leave the text a pixel or two from where it was.

#614 cached the expensive renderers and #617 corrected the fold measurement. Each removed one source of movement. Neither could remove the last one, **because the source is the rebuild itself** — a freshly built tree is not obliged to settle identically, and chasing the causes one at a time does not converge.

### 120ms → 500ms

The interval is now chosen against a **pause**, not against the gap between two keystrokes.

At 120ms the preview rebuilt several times inside one sentence and the reader watched it resettle each time. At 500ms a sentence typed without stopping is one rebuild, and it lands once they have already stopped to read it.

### What this is not

It makes the rebuild **rare**, not **stable**. The fix that removes the jitter is block-level patching — reusing the DOM of blocks whose rendered HTML did not change — after which this number can come back down.

The code comment says all of this, so the next person to see `500` does not read it as a tuned throughput value.

### Verification
- `npm test` 963 pass / 0 fail
- `npm run check` 769 files / 0 errors / 0 warnings

### DIFF
--- a/src/lib/MarkdownViewer.svelte
+++ b/src/lib/MarkdownViewer.svelte
@@ -2636,11 +2636,28 @@ import { createDocumentSession, type LoadMarkdownOptions } from './sessions/docu
 			// dropped by the revision check below, several ran at once and all of
 			// the work was done regardless.
 			//
-			// 120ms sits above the burst interval, so a word typed at speed costs
-			// one render rather than five, and below the pause someone takes to
-			// look at what they wrote — the preview catches up about an eighth of a
-			// second after the last key. Much larger reads as the preview lagging
-			// the cursor; much smaller is the 16ms case again.
+			// 500ms, and the reason is not throughput — it is that a rebuilt
+			// preview is not pixel-identical to the one it replaces.
+			//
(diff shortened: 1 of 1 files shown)


PathGao:
Closing unmerged: this was a stopgap for a problem #632 removes.

Its reasoning was "a rebuilt preview is not pixel-identical to the one it replaces, so make the rebuild rare." With block patching a keystroke touches 1–5 nodes instead of 1000–6985, so the premise is gone and the half-second lag it cost is no longer buying anything.

master keeps the 120ms from #614, whose reasoning is about cost rather than stability — each render is still an IPC round trip, a full-document comrak parse and a DOMPurify pass — and that still holds.

PR #632, opened 2026-08-12
perf(preview): replace the blocks that changed instead of the article

PathGao:
Every keystroke rebuilt the whole article: `{@html}` destroyed and re-created every element, then `renderRichContent` walked the new tree. A rebuilt tree is not obliged to settle identically, so #614 (memoise the renderers), #617 (fix the fold measurement) and #622 (lengthen the debounce) each removed one source of movement without removing the source.

This replaces only the blocks whose rendered markup changed.

### Measured — Chromium, real pipeline, one character typed mid-paragraph

| | elements re-created | apply | reference block drift |
|---|---|---|---|
| `markdown-syntax.md` | 1000 → **2** (of 1733) | 23.8 → **5.9ms** | +0.000px |
| `katex-stress.md` | 230 → **1** (of 4641) | 27.7 → **5.3ms** | +0.000px |
| `stress-test.md` | 6985 → **1** (of 7819) | 51.1 → **17.5ms** | +0.000px |

In the running app: one keystroke keeps 10 nodes and replaces 1. Inserting a line keeps 10, inserts 1, and the block below correctly moves `4:1-4:16` → `5:1-5:16`.

### What this does not claim

**The bench could not make the old rebuild path drift.** It measured +0.000px there too — on this corpus, in this browser. #614's memos make KaTeX and highlight.js output byte-identical and the samples carry no images.

So this removes the *class* of defect — 1–5 nodes touched instead of 1000–6985 — rather than a reproduced instance of it. It is not "measured the jump, fixed the jump."

### The four decisions

**Diff key: rendered markup with `data-sourcepos` stripped.** Not hashed — the string *is* the key, so one map comparison and no collision risk. `data-sourcepos` is excluded because inserting a line shifts it on every block below, which would make the whole document "changed".

**Sanitize is untouched.** `$derived(sanitizeMarkdownHtml(htmlContent))` still runs over the whole document. The patcher parses that one string into a `<template>` — an inert document, so discarded blocks never load an image — and moves nodes out of it. No string cutting, no second filter pass, so the sanitize-last argument and the `singleImplementationConvention` rule both hold as written.

**`renderRichContent` takes `roots: HTMLElement[]`** instead of one root, and runs only over changed blocks. `renderMathInElement` — the one pass #614 could not memoise, because KaTeX's auto-renderer walks the tree with its own bundled copy — is now per-block rather than per-document.

**Export forks after the detached tree.** Export passes `roots: [wrapper]`; the shared part is render → filter → detached tree, and the fork is serialise versus diff-in. Nothing renders twice.

**`data-sourcepos` is refreshed on kept nodes.** They can be stale even in an unchanged prefix — a nested container's whole child list shifts when a line is added above the heading that owns it — and split-view scroll sync reads it while the user is typing. It is an attribute no CSS rule matches: no invalidation, no layout, no paint.

### Known bad case

**Renaming a heading touches 71 blocks instead of 1** on `stress-test.md`. comrak dedupes heading ids (`objectives-24`, `-25`, …), so renaming one renumbers every later duplicate. Those blocks genuinely *are* different markup — no key can call them unchanged without lying. Still a third of the document rather than all of it.

### Two things the work turned up

**A regression the bench caught.** A patched block is not obliged to be the same height, and the fold wrapper stayed pinned to its old `--fold-content-height` for a frame while `.content-inner` spilled over the next section. `observe()` now releases the pinned height above an inserted block. A rebuilt wrapper never showed this, because it carries no inline property.

**`{@html sanitizedHtml}` had to go.** Svelte tracks its first and last node and removes everything between them on update, so a diff that removes either end breaks it silently. The document now lives in its own `<div class="markdown-blocks">`; the front-matter panel stays Svelte's.

Rewriting `previewSanitize.test.ts` to follow the chain to its new home turned up that **its old regex was matching on comments** — a false green, closed by anchoring to line start.

### Net subtraction elsewhere

Three enrichment call sites collapsed to one, and the fold observation no longer tears itself down on every keystroke.

`npm test` 996 · `npm run test:vitest` 81 · `cargo test` 145 · `npm run check` clean.

### DIFF
--- a/scripts/blockPatch.spec.ts
+++ b/scripts/blockPatch.spec.ts
@@ -1,6 +1,13 @@
 import { beforeEach, describe, expect, test } from 'vitest';
 
 import { patchPreviewBlocks } from '../src/lib/utils/blockPatch.js';
+import { asRendererLine } from '../src/lib/utils/lineCoordinates.js';
+import {
+	findAnchorElement,
+	getPreviewOffsetForSourceLine,
+	type AnchorBox,
+	type AnchorNode,
+} from '../src/lib/utils/previewAnchor.js';
 
 /**
  * The shape `processMarkdownHtml` actually produces: a heading, then a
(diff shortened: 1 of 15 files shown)


PathGao:
**Hold this one.** It breaks a cache contract that is written out in `previewAnchor.ts`:

> WHAT INVALIDATES THEM: nothing, explicitly. […] neither of which is edited in place — the preview is rebuilt wholesale by `bind:innerHTML` on every render pass, so a changed document is a new set of nodes.

`refreshSourcepos()` rewrites `data-sourcepos` in place on kept nodes. The node survives, so its WeakMap entry survives with the old parsed range, and nothing ever replaces it.

Measured with `previewAnchor` untouched — two blocks at lines 1 and 3, boxes at 0px and 100px; fill the cache, rewrite the attributes to lines 3 and 5:

| query | answers | should be |
|---|---|---|
| line 3 | 100px | 0px |
| line 5 | 300px | 100px (extrapolated past the document) |

Every edit that shifts line numbers makes it worse and it never recovers. Scroll sync, the tab's reading position and the outline's current heading all read through this.

The PR's own note said the attribute write is "no CSS rule reads it: no invalidation, no layout, no paint" — right about CSS, wrong about the memo.

A fix is being added to this branch.

PathGao:
Fixed on this branch in `ef97c84`.

`previewAnchor.ts` exports `invalidateAnchorMemos()`, which swaps both WeakMaps for fresh ones. `patchPreviewBlocks` calls it **before** it touches the tree — a patch that throws half-way through has still moved part of it, and an empty memo is a slow answer where a stale one is a wrong one.

**Whole-map, not per-node**, and `spanCache` is why. A span is a property of the *subtree*, so keeping a `.foldable-content-wrapper` while replacing a paragraph inside it leaves the wrapper's span stale **without the wrapper ever being touched**. Per-node invalidation would have to walk to the root from every mutation site, and would silently under-invalidate the first time someone added a site that skipped the walk.

It also turns out `refreshSourcepos` is not the only in-place writer — `descend` writes through `carrySourcepos` onto the kept container too. Two assignments cover both without enumerating them.

Cost: one cold lookup per render pass. Renders are debounced against a typing pause; scroll events are not, and every one after the first is warm again.

The contract comment was rewritten rather than patched, and `blockPatch.ts`'s "an attribute write no CSS rule reads" paragraph now carries the correction.

Falsified: commenting out the call turns 2 tests red with exactly the reported symptom — `expected 100 to be +0` for `ownRangeCache`, `expected undefined to be <p data-sourcepos="10:1-10:5">` for `spanCache`. Both tests drive `previewAnchor`'s public entries across a real `patchPreviewBlocks` call; neither asserts that the invalidator was called.

`npm test` 1036 · `npm run test:vitest` 89 · `npm run check` clean.

PR #645, opened 2026-08-12
feat(editor): add and remove table rows and columns without editing pipes by hand

### LINKED ISSUE #604: 建议优化表格和列表的编辑
### What problem would this solve?

1. 目前表格编辑增加一行或一列，只能手动修改代码。
2. 目前列表只能选中文本再设置编号。

### What would you like?

1. 希望能够通过右键或其他方式，方便地在表格边缘增加一行或一列。
2. 希望列表能够自动补全，比如打了`1. `以后，回车换行能自动带上`2. `，能够支持多级列表（通过`Tab`键），同时也能支持无序列表。

### What have you tried instead?

纯手动编辑。

PathGao:
Closes the row/column half of #604: adding a row or column at a table's edge without editing the pipes by hand.

Hand-tested, and the first round found four defects — the fixes are folded in below rather than left for a follow-up.

### The keys

**Cells — `Tab` / `Shift+Tab`.** With a sole collapsed caret in a table, move to the next/previous cell; `Tab` in the last cell appends a row. The caret lands at the **end** of the target cell's text. This is where every editor with table support agrees.

**Rows — `Mod+Enter` / `Mod+Shift+Enter`**, insert below and above.

These are not a new invention: Monaco already binds them to `insertLineAfter` / `insertLineBefore`, and inside a table "insert a line below" simply *is* "insert a row below". Same key, same meaning, specialized by where the caret is — the pattern this branch already uses for `Enter` and `Tab`, and why all five live in one panel group rather than among the chords. Outside a table they are handed straight back to the core action, unchanged.

An earlier revision put rows on `Mod+K R`. That was wrong twice over: awkward to press for the most frequent table edit, and a second thing to learn where an existing key already meant it.

**Columns — `Mod+K C`**, alongside the pre-existing `Mod+K T` (Insert Table). Columns are the rarer edit, so they carry the chord.

**No delete chord.** `Mod+K Shift+R` / `Mod+K Shift+C` were removed after testing: `Cmd+Shift+K` is Monaco's own `deleteLines`, one slip away, and a mis-fired destructive table edit costs far more than a mis-fired insert. The operations and their tests stay, and the actions stay registered — `addAction` without keybindings still lists them in the command palette, which is the right home for a destructive verb.

**No context-menu entry.** The viewer's three menu builders are for the *preview*; a table verb belongs to the editor, and adding one invites the same for every other editing command.

### The caret column was computed in the wrong unit

The defect hand-testing surfaced, and the one worth reading. `editAt` accumulated `widths[i] + 3`, where `widths` is **display columns** (CJK counts 2) but a Monaco column is a **UTF-16 offset** (CJK counts 1). Tabbing from `甲` in `| 甲   | 北京 | 一   |` returned column 10; `北京` starts at 9, so the caret landed *between* 北 and 京, one column off per CJK character in every preceding cell.

Fixed structurally, not arithmetically: `editAt` already computes `formatTable(table)`, so the column is now **read back out of the line the edit is about to write**. There is no second derivation left to drift — `columnWidths()` is reachable only from `formatTable`. Tests assert by slicing the produced line at the produced column, which no width-based re-derivation can satisfy.

### Chords with the modifier held

`Cmd+K T` reached the app's new-tab and `Cmd+K R` did nothing, both because the second key was pressed with Cmd still held — which is how anyone reads "Cmd+K T". VS Code registers both forms; `addAction` takes an array, so this is one extra entry.

**`Mod+K Mod+C` is deliberately not registered.** It is Monaco's `addCommentLine`, and it is not dead here: the Markdown language config declares a `blockComment`, so `LineCommentCommand` wraps the line in `<!-- -->`. Dynamic keybindings register at weight 1000 against EditorContrib's 100, so we would have won and silently deleted a working command. `Mod+K C` released is unclaimed and works. The exemption is a named entry carrying that reasoning, with a rot check.

### Alignment

Column width is `max(3, widest cell)` — the floor so `:-:` fits the column it delimits, which is Prettier's floor too. Delimiter markers are **stretched, not reset** (`:---` → `:-----`), so alignment travels with its column and inserting in front of a right-aligned column does not steal its `--:`.

`displayWidth()` counts columns, not characters. Verified: every line of a mixed CJK/ASCII table comes out at exactly equal display width. Residual visual unevenness in the editor is the CJK fallback font's advance not being exactly 2× the Latin one — no padding logic can fix that.

**Two limitations, tested rather than hidden:** ZWJ sequences are counted per code point, so `👨‍👩‍👧` measures 6 instead of 2 and over-pads its cell (fixing it wants a grapheme segmenter nothing else here needs); and a row must start with `|`, because accepting GFM's optional outer pipes would let `ls | grep foo` in prose get re-flowed into a table.

### Panel

A fifth group, `keys`, for keys whose meaning depends on where the caret is — one row per **key**, not per behaviour, since two rows advertising `Enter` would fail the chord-uniqueness check and misdescribe what is one binding. Groups now have a rule and space between them; labels use a `Context: verb · Context: verb` pattern so long-sentence locales stop wrapping.

These are nameless `addCommand`s, so `editorAction` cannot carry them — that flag means "the id is a Monaco action id". Added `editorCommand?: string` naming the handler, verified by a test that reads the component's real `addCommand` calls with keybindings evaluated by Monaco's own `KeyMod`/`KeyCode`. Also added the **completeness direction** (`BARE_NOT_ADVERTISED`) — the check that would have caught #636 shipping an unadvertised binding.

### Coverage

`npm test` 843 → **898** · `vitest` 351 · `check` 0 errors.

26 languages. `vi`, `ro`, `hu`, `sk`, `el`, `fi` are worth a native reading; the shortened labels should travel better than the sentences they replaced.

Closes #604


PathGao:
Hand-testing found four things; **hold this until they land**.

**1. The caret column is computed in the wrong unit.** `editAt()` accumulates `widths[i] + 3`, but `widths` is display columns (CJK counts 2) while a Monaco column is a UTF-16 offset (CJK counts 1). Tabbing from `甲` in `| 甲   | 北京 | 一   |` returns column 10; `北京` starts at 9, so the caret lands *between* 北 and 京. Every CJK character in a preceding cell adds one more column of error. Same defect class this branch was written against — two units answering one question with nothing forcing agreement. The caret should also land at the **end** of the target cell, not the start.

**2. The `Mod+K` chords do not work the way they are pressed.** `Cmd+K T` reached the app's new-tab and `Cmd+K R` did nothing, both explained by the second key being pressed with Cmd still held — which is how anyone reads "Cmd+K T". VS Code registers both `Cmd+K T` and `Cmd+K Cmd+T` for this reason; we register only the released form.

**3. The two delete chords are being dropped.** `Cmd+Shift+K` is Monaco's own delete-line, one slip from `Cmd+K Shift+R`, and a mis-fired destructive table edit costs far more than a mis-fired insert. The operations and their tests stay; only the bindings go.

**4. The shortcuts panel groups run together** — headings sit flush against the previous group's last row.

Not a defect: the column padding is correct. Measured every line of a mixed CJK/ASCII table at exactly 32 display columns. Residual visual unevenness is the editor's CJK fallback font not being exactly 2× the Latin advance, which no padding logic can fix.

PR #647, opened 2026-08-12
fix(shortcuts): stop the destructive chords answering to modifiers they never declared

PathGao:
`Mod+W` in the document-level `handleKeyDown` tested the platform modifier and the key, but not the other modifiers — so `Shift+Cmd+W` and `Alt+Cmd+W` also closed the file. Its neighbours guard theirs.

Auditing it found the same shape in **twelve** branches, not two.

### What the handler was actually answering

`scripts/keymapHarness.ts` already fuzzes the full Shift/Alt cross product against the real extracted handler, so this is measured rather than read:

**macOS: 123 chords answered. The registry advertises 20.** After the fix: **51**, every one advertised or explained.

Three of the strays are destructive, and one was not in the original report:

| branch | strays |
|---|---|
| `Mod+W` → `closeFile` | `Shift+Mod+W`, `Alt+Mod+W`, `Shift+Alt+Mod+W` |
| `Mod+Q` → window close | the same three — including `Shift+Cmd+Q`, which is macOS **Log Out** |
| `Mod+F4` → `closeFile` | the same three, a third path to the same irreversible action |

And one collision worth naming: `Mod+E` also answered `Shift+Mod+E`, which is `fmt-inline-code` *inside* the editor. One chord, two meanings, decided by which pane had focus.

### The three decisions

**Stray variants do nothing, and are not `preventDefault`'d.** The handler is a flat list of `if`s with no fallthrough chain to reach, and these chords have no second meaning in Markpad. Leaving them both unhandled and unprevented is the point: `Shift+Cmd+W` and `Alt+Cmd+W` are window-manager gestures, so the OS gets them back rather than the app answering with the one irreversible thing it can do.

**One shared guard, not fourteen.** Three named shapes — `mod`, `modShift`, `modAlt` — so a branch added tomorrow inherits the guard by reaching for `mod` instead of remembering three negations. It is the neighbours' guard spelled once; the correct branches mostly shrink.

**But a helper only helps branches that reach for it**, so the inheritance mechanism is a test rather than a convention. `shortcutRegistry.test.ts` already asked *"is every command the handler can reach advertised?"* — which `closeFile` passed while four undocumented chords ran it. The same question now runs **per chord**, which is the granularity the bug lived at. A future branch that forgets its guard fails there and has to either advertise what it bound or write down why not. Eight genuinely-unadvertised chords (`Mod+F4`, `Mod+PageUp/Down`, `Cmd+Alt+arrows`, ISO backslash, `Meta+Tab`) are now named with reasons instead of being invisible.

### One guard would have broken real behaviour

`key === '+'` **is** the shifted `=` on US/UK layouts, so applying the blanket `mod` shape to zoom-in would have deleted the `+` spelling and killed `Cmd+Shift+=` for everyone who zooms that way. That branch guards per spelling instead: `!altKey && (key === '+' || (key === '=' && !shiftKey))`.

Tab cycling is the other carve-out — there Shift is the **argument** (direction), not part of the chord's identity — so it guards Alt only. Both are commented in place.

### Falsification

- *a close chord wearing an extra modifier destroys nothing* — red on master with `macOS: Ctrl+Alt+W runs closeFile`. Asserts no chord matching `/\b(Shift|Alt)\b/` reaches `closeFile` or `getCurrentWindow`, plus a floor of ≥3 plain close chords so it cannot pass vacuously against an empty keymap.
- *every chord the document handler answers is advertised, or consciously not* — red on master with a 72-entry diff on macOS alone. Now empty.

Both drive the real extracted handler, not a description of it.

`npm test` 843 → **845** · `vitest` 351 · `check` 0 errors.

Five existing tests pinned the old branch text and were retargeted to the new spelling — they went red for that reason, not for a behaviour change.

### DIFF
--- a/scripts/jumpToSelectedFragment.test.ts
+++ b/scripts/jumpToSelectedFragment.test.ts
@@ -440,7 +440,7 @@ test('one function owns what ⌘E means, and every entry point uses it', () => {
 	// here, so the chord cannot mean one thing with the caret in the editor and
 	// another with it in the preview — which is what
 	// `formatShortcutKeymap.test.ts` pins from the other side.
-	assert.match(viewerSource, /if \(cmdOrCtrl && key === 'e'\) \{[\s\S]{0,600}?toggleEditView\(\)/);
+	assert.match(viewerSource, /if \(mod && key === 'e'\) \{[\s\S]{0,600}?toggleEditView\(\)/);
 	assert.equal(
 		(viewerSource.match(/ontoggleEdit=\{\(\) => toggleEditView\(\)\}/g) ?? []).length,
 		3,
--- a/scripts/previewWidth.test.ts
+++ b/scripts/previewWidth.test.ts
@@ -88,7 +88,9 @@ test('preview width keyboard shortcuts avoid editable controls and app modals', 
 	assert.match(viewerSource, /function canUsePreviewWidthShortcut\(target: EventTarget \| null, isSplit: boolean\)/);
 	assert.match(viewerSource, /showSettings \|\| modalState\.show \|\| promptModal\.show \|\| showHome/);
 	assert.match(viewerSource, /closest\('input, textarea, select, \[contenteditable="true"\], \[role="textbox"\]'\)/);
-	assert.match(viewerSource, /cmdOrCtrl && e\.altKey && !e\.shiftKey && \(code === 'BracketLeft' \|\| code === 'BracketRight'\)/);
+	// `modAlt` is the handler's name for "the platform modifier and Alt, and
+	// nothing else" — the chord these two keys are advertised under.
+	assert.match(viewerSource, /modAlt && \(code === 'BracketLeft' \|\| code === 'BracketRight'\)/);
 	assert.match(viewerSource, /settings\.previewFullWidth = false/);
 	assert.match(viewerSource, /settings\.previewMaxWidth = adjustPreviewMaxWidth\(settings\.previewMaxWidth, code === 'BracketLeft' \? -40 : 40\)/);
 });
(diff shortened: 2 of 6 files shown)


PathGao:
Added `Mod+F4` platform gating on top.

Hand-testing found `Cmd+F4` did nothing on macOS. That turned out not to be this branch's guard — the branch never should have been cross-platform. `Ctrl+F4` closes a document window on Windows and Linux, an MDI-era convention that is still live; macOS has no equivalent, and with the default "Use F1, F2, etc. as standard function keys" setting *off*, the key does not reach the app at all. So on macOS it was a third route to an irreversible action that was simultaneously non-conventional and near-unpressable.

The test being applied is worth naming, because a sibling chord looks similar and stays: **"does anyone on this platform actually press this?"**, not "does it carry an unusual modifier?". `Cmd+Shift+=` keeps firing everywhere — it is not a platform convention at all, just how `+` is typed on every layout.

**Gated on `platformOf`, not a bare `!== 'macos'`.** The latter would have been legal — `singleImplementationConvention.test.ts` explicitly permits a bare `=== 'macos'` guarding an `&&` — but `osType` is `'unknown'` until the Rust round trip returns, and `!== 'macos'` resolves that startup window to *fire the close branch*, on a Mac. That is precisely the default #646 landed the helper to delete.

One harness change was load-bearing rather than cosmetic: free identifiers inside the extracted `handleKeyDown` resolve to a recording stub, and a stub's result `=== 'windows'` is false on every platform — so the gated branch would have been reported as firing everywhere and the new test would have failed against correct source. The real `platformOf` is now in scope.

macOS answered chords 51 → **49**; Windows/Linux unchanged at 53. The drop is 2 rather than 1 because both `Ctrl+F4` and `Meta+F4` reached it. Neither count assertion needed re-baselining — both are floors against empty iteration, not snapshots.

One refinement to my own reasoning, for the record: macOS reachability is not zero. With the F-key setting **on**, or fn held, `Cmd+F4` does reach the webview — the pre-fix harness confirms the branch would have run. It is gated behind a non-default setting plus a chord nobody on the platform has reason to press. The convention argument carries the change by itself.

`npm test` 845 → **846** · `vitest` 351 · `check` 0 errors.

Issue #207, opened 2026-07-11
编辑区右键菜单粘贴无效


StrollStars:
**Describe the bug**

**To Reproduce**
1. 在程序外部或内部复制任意文本
2. 右击编辑区点击粘贴菜单项
3. 没能粘贴出复制的内容

**Expected behavior**
[A clear and concise description of what you expected to happen.]
能够把复制的内内容粘贴出来。

**Screenshots**
[If applicable, add screenshots to help explain your problem.]

**Device**
 - OS: Windows 10 LTSC 21H2
 - Version: 19044.7417
 - App Version: v2.6.11

**Additional context**

PathGao:
Addressed by #266 (v2.6.14). The cause was the app's own context menu: it called `preventDefault()` before checking whether the click was inside the editor, and its full-screen overlay then sat on top of Monaco's native menu — so clicking Paste hit the overlay. The handler now returns early inside the editor and Monaco's own menu is no longer covered.

I can't verify the rest from macOS: Monaco's Paste goes through `navigator.clipboard.readText()`, and whether WebView2 permits that read is a Windows question. Could you try again on v2.6.14 or later and say whether Paste works now?

PathGao:
The menu half is fixed, in v2.6.14 rather than v2.7.0 (#266): Markpad's own context menu called `preventDefault()` before checking whether the click was inside the editor, and its full-viewport overlay then covered Monaco's native menu, so clicking Paste hit the overlay. The handler now returns early inside the editor. I still can't verify the clipboard half from macOS — Monaco's Paste goes through `navigator.clipboard.readText()`, and whether WebView2 allows that read is a Windows question. Could you retest and say whether Paste works now?

PathGao:
Fixed in #548, and this closed with it. @Bluewind0303 — the next release after
2.7.2 will have it; no need to retest a build, the cause turned out to be
something that could be read off the source once we knew where to look.

Recording what it was, because the diagnosis went one layer deeper than the
report and the first fix was wrong.

## Two separate bugs wore the same symptom

**The menu was covered.** Markpad's own context menu called `preventDefault()`
before checking whether the click was inside the editor, and its full-viewport
overlay then sat on top of Monaco's native menu, so clicking Paste hit the
overlay. That was #266, in v2.6.14.

**The clipboard was refused.** With the menu no longer covered, Paste reached
Monaco's own implementation, which reads the clipboard through the webview.
[...]

PR #667, opened 2026-08-14
fix(menu): the editor's context menu gets the entries that are about the document

### LINKED ISSUE #97: The options of the right-click pop-up window
**Describe the bug**
In the Split View mode, when right-clicking in the editor area and the view area respectively, the options of the context menu that pops up are inconsistent.

**To Reproduce**
1. Switch to Split View mode
2. Right-click in the editor area
3. Observe the right-click menu options
4. Right-click in the view area
5. Compare the options of the two popped-up menus

**Expected behavior**
In the Split View mode, the right-click menus of the editor area and the view area should remain consistent.

**Screenshots**

Image

**Device**
 - OS: win 10 pro
 - Version: 22h2
 - App Version : 2.6.4

**Additional context**
This situation may occur in a Chinese environment. Other languages may also appear, such as English.

Image

Image

PathGao:
## What this is

Closes #97, reported against 2.6.4.

In split view the editor's context menu and the viewer's are different lists.
This adds the three entries that were different for no reason to do with where
the click landed.

## Mechanism

The two menus, as they were:

```
editor                       viewer
──────────────────────       ────────────────────────────────
Cut / Copy / Paste           Open in New Tab   (on a link)
─                            Copy Reference    (on a heading)
Command Palette      F1      Save Image As     (on media)
Change All Occurrences ⌘F2   Copy             (with a selection)
                             Select All
                             ─
                             Open File Location
                             Edit
                             ─
                             Close File
```

Most of that difference is the panes being different. Three entries are not:
**Select All** acts on the text either pane is showing, and **Open File
Location** and **Close File** act on the document that both of them are. Those
three are now in both.

**The note on the issue is out of date, which is what makes this fixable.** It
says the editor menu is Monaco's built-in and therefore different — true in
April. #207 replaced it with one this app draws, because Monaco's Paste reads
the clipboard through the webview and cannot work here. Both menus are ours now,
so the scope proposed in the thread (sync Monaco's UI language, add some viewer
actions) is no longer what the problem needs.

**Select All is a new export on the editor, not an id through
`runEditorAction`.** That resolver is:

```ts
editor.getAction(actionId)?.run();
```

and `getAction` is `this._actions.get(id)`, a map `codeEditorWidget` fills from
the editor-action registry alone. `editor.action.selectAll` is not in it — it is
registered by `registerCommand` as a `MultiCommand` (`editorExtensions.js`), and
`editor.trigger` reaches that kind only by falling through to the command
service. So the id resolves to `undefined`, the `?.` swallows it, and the menu
entry would have appeared and done nothing.

## Scope

- **Not identical lists, and there is a test saying so.** Cut and Paste need
  somewhere to type; the two Monaco entries need Monaco; Open in New Tab, Copy
  Reference, Save Image As and Edit each need a rendered element to have been
  clicked. An assertion holds each of those to one menu, so that "make them
  consistent" does not become "make them the same" in a later pass.
- **Copy is in both already** and is left alone — the editor's copies the
  selection through the editor, the viewer's through the DOM selection.
- **`runEditorAction` is unchanged**, and its silent miss is worth knowing
  about: any future menu entry given a command id rather than an action id will
  appear and do nothing, with nothing to say so. Making it fall through to
  `editor.trigger` would fix the class — `trigger` tries `getAction` first, so
  existing callers would not change — but that is fifteen call sites' worth of
  dispatch change for one menu entry, so it is recorded rather than done.
- **`editorPane`'s type is a hand-written mirror** of the editor's exports, so
  it needed the new member. `svelte-check` caught that, which is the mirror
  working.

## Tests

`scripts/contextMenuParity.test.ts`. Both menus are built inside a 4,700-line
component, so these assert against the source; what they establish is that each
label appears in the right builder, not what happens when it is clicked.

Reverted three ways:

| revert | result |
|---|---|
| drop Select All from the editor menu | 1 pass / 2 fail |
| send Select All through `runEditorAction` | 2 pass / 1 fail |
| drop Close File from the editor menu | 2 pass / 1 fail |

## Verification

```
npm audit            0 vulnerabilities
npm run check        798 files, 0 errors, 0 warnings
npm test             918 pass, 0 fail
npm run test:vitest  41 files, 365 pass
```

No new strings: `menu.selectAll`, `menu.openLocation` and `menu.closeFile` are
already in all 26 locales, which is what made this three lines of menu.

**Not checked by hand yet** — that the three entries appear in the editor's menu
in split view and each does what it says, Select All included, since that is the
one whose failure mode was silence. I will report it here rather than leave the
gap implied.

### DIFF
--- a/scripts/contextMenuParity.test.ts
+++ b/scripts/contextMenuParity.test.ts
@@ -1,0 +1,70 @@
+import assert from 'node:assert/strict';
+import test from 'node:test';
+
+import { functionSource, readSource } from './sourceTree.js';
+
+/*
+ * Issue #97: in split view the two context menus are different lists.
+ *
+ * Some of that is the panes being different — Cut and Paste need somewhere to
+ * type, the two Monaco entries need Monaco, and the viewer's own entries each
+ * need a rendered element to have been clicked. What is not is the three
+ * actions that have nothing to do with where the click landed: Select All acts
+ * on the text either pane is showing, and the two file actions act on the
+ * document that both of them are. Those were in the viewer's menu only.
+ *
+ * The maintainer's note on the issue — that the editor menu is Monaco's
+ * built-in and therefore different — was true when it was written and is not
+ * now: #207 replaced it with one this app draws, because Monaco's Paste reads
+ * the clipboard through the webview and cannot work here. Parity is reachable
+ * because both menus are ours.
+ *
+ * Both menus are built inside a 4,700-line component, so this asserts against
+ * the source. What it establishes is that each shared label appears in both
+ * builders — not what happens when it is clicked.
+ */
+
+const viewer = readSource('src/lib/MarkdownViewer.svelte');
+const editorMenu = functionSource(viewer, 'showEditorContextMenu');
+const documentMenu = functionSource(viewer, 'handleContextMenu');
+
+test('the actions that are about the document are in both menus', () => {
+	for (const key of ['menu.selectAll', 'menu.openLocation', 'menu.closeFile']) {
+		assert.match(editorMenu, new RegExp(key.replace('.', '\\.')), `${key} is missing from the editor menu`);
+		assert.match(documentMenu, new RegExp(key.replace('.', '\\.')), `${key} is missing from the viewer menu`);
+	}
+});
+
+test('the actions that are about the pane stay in one menu', () => {
+	// Asserted so that "make them consistent" does not turn into "make them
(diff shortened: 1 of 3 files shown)


PathGao:
Closing: the three entries this adds are all reachable already (⌘A, ⌘W, the title bar), so it buys discoverability of things that are not hard to find while making the editor's context menu half again as long for everyone. Reasoning written up on #97, which is closed with it.

Issue #722, opened 2026-08-26
Do not clear the temporary files


ProudRabbit:
### Markpad version

2.7.5

### How did you install it?

Windows — portable (Markpad_….exe)

### OS and version

Windows 10 22H2

### What happens, and what you expected instead

每次我保存文件(<kbd>Ctrl+s</kbd>)时，MarkPad 都会创建一个临时文件。当我关闭文档和 MarkPad 时，这些临时文件不会被删除。

```
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:18 .验证结果.md.markpad-tmp-23436-1787728728439865400-4
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:18 .验证结果.md.markpad-tmp-23436-1787728730336559800-5
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:18 .验证结果.md.markpad-tmp-23436-1787728733734697700-6
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 14:45 .验证结果.md.markpad-tmp-27716-1787726751459097400-4
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:25 .验证结果.md.markpad-tmp-33956-1787729117209943400-3
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:26 .验证结果.md.markpad-tmp-33956-1787729161741606400-6
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:26 .验证结果.md.markpad-tmp-33956-1787729168309099600-7
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:24 .验证结果.md.markpad-tmp-34932-1787729079878793300-3
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:24 .验证结果.md.markpad-tmp-34932-1787729081838730000-4
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:24 .验证结果.md.markpad-tmp-34932-1787729085054630400-5
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:24 .验证结果.md.markpad-tmp-34932-1787729085103535000-6
-rw-r--r-- 1 xxxxxx 1049089      0 Aug 26 15:22 .验证结果.md.markpad-tmp-35040-1787728960314673700-4
[...]

Issue #562, opened 2026-08-10
Packaging: one fix removes half of this list, and the rest is Linux


PathGao:
@alecdotdev @LargeModGames

Markpad ships eleven artifacts across six channels, and the packaging issues open against them are not eleven separate problems. This is what they reduce to, what has already been fixed without needing anyone's decision, and the three things that are genuinely yours to decide.

Long evidence is folded into the `<details>` blocks. The top level is the part worth reading.

---

## The whole problem in one table

Every channel has to answer three questions:

| channel | installs | can update itself | trusted by the OS |
|---|:--:|:--:|:--:|
| Windows `*-setup.exe` | ✅ | ✅ | ❌ #334 #466 |
| Windows portable `.exe` | ✅ | ❌ | ❌ |
| Chocolatey | ✅ | ✅ | ❌ |
| macOS `.dmg` | ✅ | ✅ | ❌ #209 #505 #557 |
| Linux AppImage | ⚠️ #182 #498 | ✅ | — |
| Linux `.deb` | ✅ | ❌ | — |
| Linux `.rpm` | ✅ | ❌ | — |
| Linux snap | ✅ | 💀 dead since June | — |

**The third column is one cause, not four.** Every ❌ there is the same missing thing — a code-signing identity. It accounts for five open issues and needs zero lines of code.

**Once that column is green, everything left is Linux.** The only exception in the list is #84 (custom install path on Windows), an ordinary feature request rather than part of this.

### How many people are behind each row

GitHub publishes per-asset download counts. v2.7.2, three days after release:

```
Windows  x64 installer     540  ┐
         x64 portable      245  │  818 = 91%
         arm64 installer    17  │
         arm64 portable     16  ┘
macOS    universal.dmg      37  =  4%
Linux    AppImage           26  ┐
         .deb               15  │   44 =  5%
         .rpm                3  ┘
```

Chocolatey's feed reports 6–22 installs per version. The Snap Store publishes nothing — that is structural, and the snap section below explains why. The ratio holds across v2.7.0 and v2.7.1 too, so it is not an artefact of a three-day window.

These are downloads, not users, and they say nothing about who is already installed and updating in place. But the shape is stark enough to order the work:

- **Windows signing reaches twenty times more downloads than macOS signing.** macOS has the louder issues; Windows has the volume.
- **Linux is 5% across all four channels.** #498 was a genuinely broken artefact and worth the fix, but section 2 should not cost more than section 1.
[...]

Issue #466, opened 2026-08-05
Windows portable release flagged by Microsoft VirusTotal heuristic detection


cindyyoyo:
**Describe the bug**

The Windows portable release of Markpad 2.7.0 appears to trigger false-positive malware detections on VirusTotal.

**To Reproduce**

1. Download the Markpad 2.7.0 portable Windows x64 release from the official GitHub release page.
2. Verify that the SHA-256 checksum matches the published checksum.
3. Scan the executable with VirusTotal.
4. Observe the following detections:

   * Microsoft: `Trojan:Win32/Wacatac.B!ml`
   * Trapmine: `Suspicious.low.ml.score`

**Expected behavior**

The official release binary should ideally not trigger false-positive malware detections.

**Actual behavior**

VirusTotal reports 2/70 heuristic/ML-based detections for the official portable Windows release.

**Additional context**

VirusTotal report:
https://www.virustotal.com/gui/file/ed58d83b7342dcbc58352202a472d719747d5070b373dd537a8c997e38876646

The executable is not digitally signed, so this may be related to reputation or heuristic detection.

The SHA-256 checksum matches the one published with the release.

I wanted to report this so other users are not discouraged from using the portable version due to antivirus warnings.

Thanks for maintaining Markpad.

PathGao:
Not malicious, and I think it is worth naming exactly what trips it rather than leaving it at "false positive" — the pattern is real, it just belongs to the uninstaller.

@alecdotdev this is the same root cause you diagnosed in #334 ("a false positive caused by the custom uninstall script"). It is still there, and it now shows on the **portable** exe because the portable and the installer are the same binary.

## What the scanner is reacting to

`src-tauri/src/setup.rs`, the uninstall path:

```
:898   writes  %TEMP%\markpad_uninstall_<suffix>.bat
:920   writes a .vbs:  CreateObject("Wscript.Shell").Run "<bat>", 0, False
                                                          ↑ 0 = hidden window
:922                   fso.DeleteFile WScript.ScriptFullName    ← the script deletes itself
:956   launches it via C:\Windows\System32\wscript.exe
:843   Command::new(&target_exe).spawn()
```

[...]

PR #753, opened 2026-09-04
Sign the Windows executables through SignPath

PathGao:
## What this is

Windows Authenticode signing through SignPath Foundation, using the four strings @alecdotdev posted in #562. Ref #562, #334, #466, #722.

## Mechanism

One signing request covers all four executables — SignPath signs the contents of one Actions artifact and an approver releases it by hand — but x64 and arm64 are two matrix entries, and each uploaded straight to the release page. So the signing is a job of its own: the matrix hands its build over as an artifact, `sign-windows` collects both, submits one request, and is the only place a Windows file reaches the release. An approval that never comes leaves the draft without Windows assets rather than with unsigned ones.

Authenticode rewrites the installer's bytes, and `*-setup.exe.sig` is a minisign signature over the pre-signing ones. Shipping that makes every Windows install reject its next update — the same defect the repacked AppImage had, fixed the same way: re-sign from the signed file, then let `generate-update-feed` read the signatures back off the release.

## Scope

Without `SIGNPATH_API_TOKEN` the job uploads what the matrix built, unsigned, and the release notes keep the SmartScreen note — both keyed on that one secret, so a release cannot claim to be signed when it is not. Same shape as the macOS certificate, for the reason #294 was closed: a release must not be blocked on credentials nobody has configured yet.

The four non-secret values are in the job rather than in secrets, and RELEASING.md gets a one-time setup step 8 saying so, plus what the SignPath side needs — the GitHub App, the trusted build system, the artifact configuration, an approver.

Not here: signing the copy of Markpad an NSIS installer unpacks. Nothing can reach inside an installer after it is built, so that would be a second build pass and a second approval, and what people report — SmartScreen on the download, the heuristic on the portable `.exe` — is on the files this signs.

## Tests

Four in `scripts/releaseWorkflow.test.ts`, all matching workflow text, which is what a workflow makes checkable: no `gh release upload` of an `.exe` in the matrix; one signing request per release over four collected executables; the `.sig` regenerated after the signed executables land and before the upload, with `generate-update-feed` waiting on the job; and the upload not gated on the token while every SignPath step is. The existing "RELEASING.md names the signing secrets the workflow reads" now reads the new job too.

Each was falsified: putting a release upload back in the matrix, asking for two files instead of four, moving the re-sign ahead of the signed copy, gating the upload on the token, dropping `sign-windows` from the feed's `needs`, and renaming the secret in RELEASING.md each turn exactly one of them red.

## Verification

macOS 26.6.2, Node 24.16.0.

```
npm audit
npm run check
npm test
npx vitest run
```

0 vulnerabilities, 831 files / 0 errors, 1033 pass (was 1029), 434 pass.

Not verified: the release path itself. `build.yml` is `workflow_dispatch` only, so nothing here has run — the first proof is 2.7.7, and it is also the first time the SignPath side is exercised end to end. `cargo test` not run; no Rust in the diff.


PathGao:
@alecdotdev have you got the sign?

PathGao:
I may sign an apple developer today so probably our macos version will also be signed by me

alecdotdev:
@PathGao no response yet. let me know if you do go through with the apple dev part and if theres any secrets id need to add

PathGao:
> @PathGao no response yet. let me know if you do go through with the apple dev part and if theres any secrets id need to add

done. next time when you release，macos version will be signed by apple.

PathGao:
@alecdotdev on the SignPath wait: it has been at least nine days since you said the organization was under review, and Markpad still isn't on signpath.org/projects. SignPath publishes no review time for the Foundation programme — other projects report a few weeks — so this is not yet unusual, but it is worth a nudge.

The review lives in their inbox, not here: mail support@signpath.io with the project name, the repository URL and the date you submitted the form, and ask where the application stands. That is faster than waiting for the form to answer on its own.

Issue #668, opened 2026-08-14
PDF export prints the live window, and the print stylesheet is a list of things being subtracted back out


PathGao:
PDF export prints the live window, so anything on screen can end up on the page. #99 #103 #158 #163 #173 #232 #261 #328 #332 #359 #377 #382 #407 #411 #363 are all this — fifteen over five months, each looking like a different bug (spacing, images, theme, split width, find highlights, blank pages).

**What it is today.** `print_pdf` is one line:

```rust
// src-tauri/src/commands.rs:238
window.print().map_err(|error| error.to_string())
```

So the paper is whatever DOM is on screen — title bar, tab strip, splitter, find bar, TOC overlay, plus fold state and the current theme. `@media print` in `styles.css` is a list of things being subtracted back out, and it grows by one every time a new UI part is added. Miss one and it silently lands in someone's PDF.

The HTML export does the opposite: `buildExportArticle` re-renders from `rawContent` into a detached wrapper. That is why the two exports disagree — your own comment in `styles.css` says folds "disagreed with the HTML export, which renders every fold open".

**Which way do you want to go?**

1. **PDF renders through `buildExportArticle`, like HTML already does.** One export path, and a new UI part can never leak onto paper. Cost: printed output changes for everyone — folds all open, screen theme no longer carried, and anything people currently rely on seeing in the PDF because it was on screen.
2. **Keep `window.print()`, keep subtracting.** Zero risk today, and the `@media print` list keeps growing. #407 already had to add `syncPreviewForPrint()` because in edit mode the preview DOM was never updated — a fix that exists only to compensate for printing the live window.
3. Something in between you have in mind.

I lean 1, but it changes what people get on paper, so it is your call. Happy to do the work either way.

PathGao:
Done in #795. The PDF is rendered from the document now, so the print sheet no longer has to subtract the interface back out. Verified on macOS from both reading and edit mode.

PR #795, opened 2026-09-17
fix(export): a PDF is rendered from the document, not from the window

PathGao:
ref #668

`print_pdf` printed the live window, so the paper was whatever was on screen and `@media print` was a 333-line list of app parts being subtracted back out. Fifteen issues over five months were that list missing an entry.

The PDF route now builds the same article the HTML export writes to disk, mounts it in `#print-root`, and prints that. The print sheet reveals that element and hides the rest — structurally (`#app > :not(#print-root)`), so a part of the interface added later cannot reach a page. `syncPreviewForPrint` and the diagram re-render on screen are gone with it: the preview's DOM, fold state and theme have nothing to do with what prints.

Two things that needed deciding and are worth flagging:

- Paper is light now whatever the screen is wearing. `#print-root` carries the light palette and diagrams are drawn with the print theme. A PDF exported from a dark screen used to come out dark.
- Images are left as `asset:` URLs rather than embedded (paper does not leave the app) and the export waits for them to decode before printing, with a 3s cap.

The article comes down on `afterprint`, not when the command returns: `print_pdf` opens the platform print sheet and returns while it is still up, so clearing it then would race the sheet for the page it is about to render. A platform that never sends the event leaves it mounted and hidden until the next export replaces it.

1024 node tests, 441 vitest, `svelte-check` clean over 838 files, and a local test bundle builds. The tests that measured the old mechanism are gone; `scripts/printsExportArticle.test.ts` asserts the new one, including that the print sheet mentions no app part. **Not verified against a real print sheet on any platform** — that is the check worth doing before merging, and the thing to look at is whether the first page shows the document rather than a blank sheet.


PathGao:
Verified on macOS 27 with a local test bundle, printing the attached torture document (collapsed callout, wrapping code, wide table, KaTeX, a Mermaid diagram, a local image).

- Reading mode, dark theme: the print preview is the document on white paper — no title bar, no tab strip, no splitter. Three pages, all content present.
- Edit mode: same three pages. This used to be the blank page that `syncPreviewForPrint` existed to compensate for.
- Diagrams print light from a dark screen; maths and the table survive the page break.
- The local image prints. It is left as an `asset:` URL and the export waits for it to decode first.
- Four prints in one session all rendered correctly, so the `afterprint` teardown and the reset before each build both hold.

Find highlights were not exercised — the article is rendered from `rawContent`, so a find mark cannot exist in it.

